PR #1, opened 2018-01-10
Let python decide which path separator to use depending on OS

blueciii:
Was trying out Masonite on windows and bypassed gunicorn (no windows support) and used waitress instead then I got an error about the stylesheet not found.


josephmancuso:
yeah I haven't tested on windows yet. The bug was that the stylesheets broke on windows? This fixes it? I will test on windows soon. Either tonight or tomorrow.

Thanks for the PR!

josephmancuso:
Tests check out. Very good. Appreciate the contribution!

PR #21, opened 2018-02-23
PEP8 compliance

mapeveri:


masonite/drivers/UploadS3Driver.py
```diff
@@ -17,6 +20,9 @@ def store(self, fileitem, location=None):
         s3 = session.resource('s3')
 
         s3.meta.client.upload_file(
-            file_location, self.config.DRIVERS['s3']['bucket'], fileitem.filename)
+            file_location,
+            self.config.DRIVERS['s3']['bucket'], fileitem.filename
```
josephmancuso: Let's put that third parameter on the next line as well. So just move fileitem.filename down the the next line too.

masonite/facades/Auth.py
```diff
@@ -1,10 +1,12 @@
-''' Authentication Class '''
+"""Authentication Class"""
```
josephmancuso: Let's put a nice space there. so instead of `"""Authentication Class"""` it should be `""" Authentication Class """`

masonite/managers/Manager.py
```diff
@@ -1,12 +1,14 @@
 class Manager(object):
-
+    """
+    Manager container class
+    """
     def __init__(self, container=None):
```
josephmancuso: Let's put a space between all class docstrings and the first class method. So just put a space between line 4 and 5

masonite/managers/QueueManager.py
```diff
@@ -1,16 +1,21 @@
 from masonite.managers.Manager import Manager
 from masonite.exceptions import DriverNotFound
 
+
 class QueueManager(Manager):
-    
+    """
+    Queue manager class
+    """
```
josephmancuso: same here. space between the docstring and the first method

masonite/managers/UploadManager.py
```diff
@@ -1,8 +1,11 @@
 from masonite.managers.Manager import Manager
 from masonite.exceptions import DriverNotFound
 
-class UploadManager(Manager):
 
+class UploadManager(Manager):
+    """
+    Upload files manager class
+    """
     def create_driver(self, driver=None):
```
josephmancuso: same here. space between the docstring and the first method

masonite/packages.py
```diff
@@ -18,8 +18,9 @@ def create_or_append_config(location, name=False):
                         config_directory + '/' + file_name)
         print('\033[92mConfiguration File Created!\033[0m')
     else:
-        ## Append to the file
-        # tempfiles is a list of file handles to your temp files. Order them however you like
+        # Append to the file
+        # tempfiles is a list of file handles to your temp files.
```
josephmancuso: get rid of that period at the end.

masonite/packages.py
```diff
@@ -34,8 +35,9 @@ def append_web_routes(location):
     # import it into the web.py file
     routes_file = os.path.join(os.getcwd(), 'routes/web.py')
 
-    ## Append to the file
-    # tempfiles is a list of file handles to your temp files. Order them however you like
+    # Append to the file
+    # tempfiles is a list of file handles to your temp files.
```
josephmancuso: same with this period.

masonite/packages.py
```diff
@@ -45,12 +47,14 @@ def append_web_routes(location):
...
 
+
 def append_api_routes(location):
     # import it into the web.py file
     api_file = os.path.join(os.getcwd(), 'routes/api.py')
 
-    ## Append to the file
-    # tempfiles is a list of file handles to your temp files. Order them however you like
+    # Append to the file
+    # tempfiles is a list of file handles to your temp files.
```
josephmancuso: and this one.

masonite/providers/ApiProvider.py
```diff
@@ -1,36 +1,37 @@
-''' A ApiProvider Service Provider '''
+""" A ApiProvider Service Provider """
```
josephmancuso: This should be `An` instead of `A` so `An ApiProvider Service Provider`

masonite/providers/AppProvider.py
```diff
@@ -1,10 +1,11 @@
-''' A AppProvider Service Provider '''
+""" A AppProvider Service Provider """
```
josephmancuso: Same here. This should be `An` instead of `A`

masonite/providers/QueueProvider.py
```diff
@@ -14,4 +15,6 @@ def register(self):
         self.app.bind('QueueConfig', queue)
 
     def boot(self, QueueConfig, QueueManager):
-        self.app.bind('Queue', QueueManager(self.app).driver(QueueConfig.DRIVER))
\ No newline at end of file
+        self.app.bind(
+            'Queue', QueueManager(self.app).driver(QueueConfig.DRIVER)
```
josephmancuso: let's put these on two different lines if PEP 8 allows for it. If not we can leave it.

masonite/providers/UploadProvider.py
```diff
@@ -1,10 +1,11 @@
-''' A Mail Service Provider '''
+""" A Mail Service Provider """
```
josephmancuso: This should be `An Upload Service Provider` not `Mail`

masonite/providers/ViewProvider.py
```diff
@@ -1,7 +1,8 @@
-''' A Mail Service Provider '''
+""" A Mail Service Provider """
```
josephmancuso: This should be `A View Service Provider`.  Not `Mail`

masonite/routes.py
```diff
@@ -1,23 +1,26 @@
...
 import importlib
 import re
 from pydoc import locate
 
 from config import middleware
 
-import cgitb; cgitb.enable()
+import cgitb
+
+cgitb.enable()
```
josephmancuso: lets get rid of this import. I'm pretty sure it's not needed. as well as the cgitb.enable thing. Just delete them

masonite/validator.py
```diff
@@ -1,48 +1,49 @@
-''' Masonite Validator Module '''
+""" Masonite Validator Module """
 
 from validator import *
 
+
 class Validator(object):
-    ''' Validator Class '''
-    def __init__(self, request = None):
+    """ Validator Class """
```
josephmancuso: Let's do a space between the docstring and the first method too

josephmancuso:
Ok so I fixed the broken PR issue although I think you might have to pull from develop again to get the latest changes. The problem was that apparently travis doesn't like using environment variables with Pull requests from outside the repository (forks). So to get around that I only ran tests if the environment variable was set. We'll have to do that for now on as a work around.

In the meantime while I was doing that it looks like the changes I requested say outdated. They are not it was just because I committed to this PR :)

masonite/providers/UploadProvider.py
```diff
@@ -1,4 +1,4 @@
-""" A Mail Service Provider """
+""" A Upload Service Provider """
```
josephmancuso: This should be `""" An Upload Service Provider """` and not `""" A Upload Service Provider """`

mapeveri: Ready, check please :)

josephmancuso:
Perfect! It looks like all the tests will pass on the `develop` branch too. The issue with your pull request not passing tests is that Travis CI doesn't use the environment variables during forks and pull requests. So as a note to future contributors, if a test requires environment variables not available to pull requests, we will need to supply an if statement in the test like so:

```python
if os.getenv('SOME_SECRET'):
    
    def test_some..():
        .... code ....

    def test_other...():
        ... code ...
```

#### Note to self and future collaborators

This way it will only run those tests if the environment variables are set. The only other way is to merge it into a new branch under control by the repository and then from there run all tests which I think is too much. This solution seems simple enough. There might be an issue when someone changes code that requires an environment variable. In this case someone will have to pull in that PR locally and ensure tests before merging into the base branch

mapeveri:
Is there a .env file for the variables? Because I do not see it.

josephmancuso:
You have to create it. just make a `.env` file in the root of the project files

mapeveri:
Great, and the variables what are they?

josephmancuso:
https://github.com/MasoniteFramework/core/blob/develop/.env-example

mapeveri:
Thanks!

PR #23, opened 2018-02-23
Fix setup.py and requirements.txt

mapeveri:


setup.py
```diff
@@ -26,8 +27,8 @@
     description='The core for the Masonite ',
```
josephmancuso: Also can you get rid of this trailing whitespace at the end?

setup.py
```diff
@@ -26,8 +27,8 @@
...
     author='Joseph Mancuso',
     author_email='idmann509@gmail.com',
-    url='https://github.com/josephmancuso/masonite',  # use the URL to the github repo
+    url='https://github.com/MasoniteFramework/masonite',
     download_url='',
```
josephmancuso: Can you get rid of the download_url? seems like we're not using it.

mapeveri:
Check please!

josephmancuso:
Thanks!

PR #24, opened 2018-02-23
Add accepted file types to upload drivers #22

### LINKED ISSUE #22: Add accepted file types to upload drivers
There needs to be a way to accept specific file types. The end state of the code should look something like:

```python
def show(self, Upload):
    Upload.accept('png', 'jpg').store(request().input(file))
```

You may have to abstract this and make a BaseUploadDriver which contains this accept method and just make sure it works on the two drivers by inheriting from them. It should be really simple and should just be a check if the filetype is inside the list

Let me know if anyone has any additional questions

mapeveri:
Closes #22


.gitignore
```diff
@@ -9,4 +9,5 @@ htmlcov
 **.DS_Store
 .env
 .pytest_cache/
-storage/compiled/
\ No newline at end of file
+storage/compiled/
```
josephmancuso: What's the reasoning behind ignoring these?

masonite/drivers/BaseUploadDriver.py
```diff
@@ -0,0 +1,26 @@
...
+        self.accept_file_types = None
+
+    def accept(self, *args, **kwargs):
+        """
+        Set files to accept
+        """
+        self.accept_file_types = args
+        return self
+
+    def valid_extension(self, filename):
```
josephmancuso: let's change this to `validate_extension` to add more of an "action" word.

masonite/drivers/BaseUploadDriver.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    Class base for upload drivers
+    """
+    def __init__(self):
+        # this will accept all file types
+        self.accept_file_types = None
+
+    def accept(self, *args, **kwargs):
+        """
+        Set files to accept
```
josephmancuso: Change this to `Set file types to accept` since it's the file types we want to accept

masonite/drivers/UploadDiskDriver.py
```diff
@@ -1,17 +1,22 @@
...
+
     def __init__(self, StorageConfig, Application):
         self.config = StorageConfig
         self.appconfig = Application
 
     def store(self, fileitem, location=None):
         filename = os.path.basename(fileitem.filename)
 
+        # Check if is a valid extension
+        self.valid_extension(filename)
```
josephmancuso: Make sure this is changed to `self.validate_extension` as well

masonite/drivers/UploadS3Driver.py
```diff
@@ -12,6 +13,9 @@ def __init__(self, Upload, StorageConfig):
     def store(self, fileitem, location=None):
         file_location = self.upload.driver('disk').store(fileitem)
 
+        # Check if is a valid extension
+        self.valid_extension(fileitem.filename)
```
josephmancuso: again here

masonite/exceptions.py
```diff
@@ -1,5 +1,13 @@
 class DriverNotFound(Exception):
     pass
 
+
 class DriverLibraryNotFound(Exception):
     pass
+
+
+class ExtensionException(Exception):
```
josephmancuso: Let's change this exception to `FileTypeException` instead.

tests/test_upload_manager.py
```diff
@@ -32,5 +37,56 @@ def test_upload_manager_switches_drivers():
...
+
+
+def test_upload_manage_accept_files_error():
+    container = App()
+
+    container.bind('Application', application)
+    container.bind('StorageConfig', storage)
+    container.bind('UploadDiskDriver', UploadDiskDriver)
+
+    class ImageTest():
```
josephmancuso: Lets move this image class outside of the test functions. This way we only need 1 class for all tests instead of copying and pasting the same class. Make it DRY

masonite/drivers/BaseUploadDriver.py
```diff
@@ -0,0 +1,26 @@
+import os
+from masonite.exceptions import ExtensionException
+
+
+class BaseUploadDriver(object):
+    """
+    Class base for upload drivers
+    """
+    def __init__(self):
```
josephmancuso: This won't work either because drivers need to have their own constructor so a `BaseUploadDriver` cannot have it's own constructor. Try possibly doing something like:

```
class BaseUploadDriver(object):

    accept_file_types = None
```

instead of putting it in a constructor.

Let me know if that makes sense.

tests/test_upload_manager.py
```diff
@@ -32,5 +37,56 @@ def test_upload_manager_switches_drivers():
...
+        @property
+        def filename(self):
+            return os.getcwd() + "/tests/static/test.jpg"
+
+        @property
+        def file(self):
+            return open(self.filename, 'rb')
+
+    img = ImageTest()
+    UploadManager(container).driver('disk').accept('jpg', 'png').store(img)
```
josephmancuso: Also create a test without the `accept()` method and you'll find a bug that says `AttributeError: 'UploadDiskDriver' object has no attribute 'accept_file_types'`

.gitignore
mapeveri: The tests for sass generate a css file.

masonite/drivers/BaseUploadDriver.py
mapeveri: Ok, it's true.

masonite/drivers/BaseUploadDriver.py
mapeveri: I do not understand, why text replaced it?

masonite/exceptions.py
mapeveri: Ok!

tests/test_upload_manager.py
mapeveri: Yes, it's true. I did not realize

tests/test_upload_manager.py
mapeveri: With this test working:

def test_upload_file():
    """
    This test is responsible for checking if you upload a file correctly.
    """

    container = App()

    container.bind('Application', application)
    container.bind('StorageConfig', storage)
    container.bind('UploadDiskDriver', UploadDiskDriver)

    img = ImageTest()
    UploadManager(container).driver('disk').store(img)

masonite/drivers/BaseUploadDriver.py
mapeveri: Sorry, I read it wrong

mapeveri:
Ready, can you check again please?

tests/test_upload_manager.py
```diff
@@ -32,5 +37,68 @@ def test_upload_manager_switches_drivers():
...
+    """
+
+    container = App()
+
+    container.bind('Application', application)
+    container.bind('StorageConfig', storage)
+    container.bind('UploadDiskDriver', UploadDiskDriver)
+
+    img = ImageTest()
+    UploadManager(container).driver('disk').store(img)
```
josephmancuso: All tests need an assertion so just stick `assert` in the front like so: `assert UploadManager(container).driver('disk').store(img)`

tests/test_upload_manager.py
```diff
@@ -32,5 +37,68 @@ def test_upload_manager_switches_drivers():
...
+    """
+
+    container = App()
+
+    container.bind('Application', application)
+    container.bind('StorageConfig', storage)
+    container.bind('UploadDiskDriver', UploadDiskDriver)
+
+    img = ImageTest()
+    UploadManager(container).driver('disk').accept('jpg', 'png').store(img)
```
josephmancuso: Same thing with an assertion here

mapeveri:
Check please :)

PR #25, opened 2018-02-24
Add template caching support #19

### LINKED ISSUE #19: Add template caching support
sometimes views don't need to be loaded over and over again and can just be rendered from a cache. I should be able to cache on a per template basis. Maybe something like:

```python
def show(self):
    view('welcome').cache() # cache forever
    view('welcome').cache_for(5, 'minutes')
    view('welcome').cache_for(5, 'hours')
    view('welcome').cache_for(5, 'days')
    view('welcome').cache_for(5, 'months')
    view('welcome').cache_for(5, 'years')
```

This should ignore the template rendering entirely and just show the already rendered cached template

mapeveri:
Closes #19


config/application.py
```diff
@@ -15,6 +15,7 @@
     'masonite.providers.StartResponseProvider.StartResponseProvider',
     'masonite.providers.SassProvider.SassProvider',
     'masonite.providers.WhitenoiseProvider.WhitenoiseProvider',
+    'masonite.providers.CacheProvider.CacheProvider',
```
josephmancuso: Because wsgi is False, we can move this all the way to the bottom of the list. Service providers should be additive and shouldn't have to be injected in random places.

tests/test_view.py
```diff
@@ -105,3 +108,27 @@ def test_view_share_updates_dictionary_not_overwrite():
...
+    """
+
+    container = App()
+
+    view = View().load_container(container)
+    view.share({'test': 'test'})
+
+    container.bind('CacheConfig', cache)
+    container.bind('CacheDiskDriver', CacheDiskDriver)
+    container.bind('CacheManager', CacheManager())
```
josephmancuso: The reason you are getting a Key Error for `Cache` is because you never loaded it into the container here. This section should look like:

```python
    container.bind('CacheConfig', cache)
    container.bind('CacheDiskDriver', CacheDiskDriver)
    container.bind('CacheManager', CacheManager(container))
    container.bind('Application', application)
    container.bind('Cache', container.make('CacheManager').driver('disk'))
    container.bind('View', view.render)
```

Notice the Application addition there since your `CacheDiskDriver` requires the `Application`. When you are building a container from scratch like that you'll need to add everything you need manually

masonite/drivers/CacheDiskDriver.py
```diff
@@ -0,0 +1,30 @@
...
+        self.appconfig = Application
+
+    def store(self, template, location=None):
+        if not location:
+            location = self.config.DRIVERS['disk']['location']
+
+        location += '/'
+
+        # Generate file cache
+        path_filename = os.path.basename(template)
```
josephmancuso: This should be `path_filename = os.path.join('resources/templates', os.path.basename(template))` because all templates are in the `resources/templates` directory. This should make the tests pass. Only problem is that the cached template is:

`{{ test }}`

which you'll see when you run the tests. It should be the rendered template meaning it should not have any Jinja2 templating in it.

josephmancuso:
To get it your `__create_cache_template()` should look something like:

```
    def __create_cache_template(self, template):
        """
        Save in the cache the template
        """
        template = template + ":" + str(time.time()) + '.html'
        self.file_cache = self.container.make('Cache').store(template, self.rendered_template, 'bootstrap/cache')
```

and your `CacheDiskDriver` should look like:

```
    def store(self, key, value, location=None):
        if not location:
            location = self.config.DRIVERS['disk']['location']

        location += '/'

        open(os.path.join(location, key), 'w').write(value)

        return key
```

masonite/view.py
```diff
@@ -12,27 +15,129 @@ def view(template='index', dictionary={}):
...
+        else:
+            return False
+
+    def __expired_cache(self):
+        """
+        Check if cache is expired
+        """
+
+        # It's the first time
+        if not self.file_cache:
```
josephmancuso: This `self.file_cache` needs to change. This doesn't work if the view is only rendered once and only works because the view renders 2 times in the test. For example do this in test:

```python
    # The first render normally
    # view('test').cache_for(5, 'minutes')

    # Second time render from cache
    assert view('test').cache_for(5, 'minutes').rendered_template == 'test'
```

comment out the first one and the template doesn't cache anymore. The expired cache needs to look inside the cache directory for the cached template and not an attribute on the class

josephmancuso:
Ok So you may be making it slightly more complicated. You really just need this:

```
render():
    if a_cached_template_exists():
        return cached_template()
```

That's really it. You can break it up into multiple methods if you like but that's what it really comes down to.

The best we to really do this is to set a timestamp on the file so something like: `test.8376848r44.html` which really means: `test.5_minutes_in_the_future.html`

Now the `a_cached_template_exists()` should look something like:

```
def a_cached_template_exists(self, template_name):
    find_template = glob.glob('bootstrap/cache/resources/templates/test*')
    if find_template:
        # a cached template exists but we need to check the timestamp on it

        if os.path.splitext(find_template[0]).split(':')[1] > time.time() # gets the timestamp
            return find_template
   return False
```

masonite/view.py
```diff
@@ -51,3 +75,120 @@ def extend(self):
...
+
+        if self.__is_expired_cache():
+            self.__create_cache_template(self.template)
+        return self
+
+    # --------
+    # Privates
+    # --------
+
+    @property
```
josephmancuso: This should not be a property. It works when it is not a property and just a normal method. Is there a specific reason this is a property?

this will cause a headache because it is the only property and it should be a method. This will make maintenance harder because it's harder to debug a parenthesis issue. Especially because its valid when it is missing but has a different behavior than normal.

tests/test_view.py
```diff
@@ -26,7 +29,7 @@ def test_view_extends_without_dictionary_parameters():
 
     view = container.make('View')
 
-    assert view('test') == 'test'
+    assert view('test').rendered_template == 'test'
 
 
 def test_render_from_container_as_view_class():
```
josephmancuso: Ok so the problem of your tests not passing and because you are not loading the container into any of the other `View()` classes. Look at line 25 in this file. This needs to be: `view = View().load_container(container)`

It should be like that in all tests (and the ViewProvider but I think you have it there)

masonite/view.py
```diff
@@ -51,3 +75,120 @@ def extend(self):
...
+        """
+
+        # If is forever
+        if self.cache_time is None and self.cache_time is None:
+            return False
+
+        # By seconds
+        cache_type = self.cache_type.lower()
+        calc = 0
+        if cache_type == "minutes":
```
josephmancuso: make sure seconds is in here too. Also to make it more developer friendly, sometimes the developer will want to only do `1 minute` or `1 second` or `1 hour` and not `1 minutes`, `1 seconds` or `1 hours` so what we can do is this:

```python
           elif cache_type == "hours" or 'hour':
              calc = 60 * 60
          elif cache_type == "days" or 'day':
              calc = 60 * 60 * 60
          elif cache_type == "months" or 'month':
              calc = 60 * 60 * 60 * 60
          elif cache_type == "years" or 'year':
              calc = 60 * 60 * 60 * 60 * 60
```

for all of them.

masonite/view.py
```diff
@@ -51,3 +75,120 @@ def extend(self):
...
+            template + ":" + str(time.time()) + '.html',
+            self.rendered_template
+        )
+
+    def __cached_template_exists(self):
+        """
+        Check if the cache template exists
+        """
+
+        path = self.__get_path_cache
```
josephmancuso: since we are removing the `@property` then that means that all instances of `self.__get_path_cache` should be `self.__get_path_cache()`. Make sure they are a callable

masonite/view.py
```diff
@@ -12,27 +16,47 @@ def view(template='index', dictionary={}):
...
+        return self
+
     def render(self, template, dictionary={}):
         self.dictionary.update(dictionary)
 
+        # Load template
+        self.template = template
+
+        # Check if use cache and return template from cache if exists
+        if self.__cached_template_exists() and not self.__is_expired_cache():
```
josephmancuso: the `and not self.__is_expired_cache()` logic is reverse. Currently the template will keep on rendering the same cached template over and over and won't actually render and cache the new template when it expires but for some reason switching it to `and self.__is_expired_cache()` made it work like it should.

masonite/view.py
mapeveri: Ok!

tests/test_view.py
mapeveri: With this it was solved

masonite/view.py
mapeveri: Perfect!

masonite/view.py
mapeveri: Yes!

masonite/view.py
mapeveri: The name of the function is ok, the problem is that the function is returned upside down. Is not it better to change the return?

josephmancuso: yeah change the return values

josephmancuso: the logic of `and not self.__is_expired_cache()` looks good you just have to flip the return values

mapeveri: Great!

tests/test_view.py
```diff
@@ -105,3 +146,26 @@ def test_view_share_updates_dictionary_not_overwrite():
...
+    container.bind('CacheConfig', cache)
+    container.bind('CacheDiskDriver', CacheDiskDriver)
+    container.bind('CacheManager', CacheManager(container))
+    container.bind('Application', container)
+    container.bind('Cache', container.make('CacheManager').driver('disk'))
+    container.bind('View', view.render)
+
+    view = container.make('View')
+
+    assert view(
```
josephmancuso: So the test case here which is why there is the bug is the test case should be:

```python
    import time
    assert view(
        'test_cache', {'test': 'test'}
    ).cache_for(1, 'second').rendered_template == 'test'
    
    assert open(glob.glob('bootstrap/cache/test_cache*')[0]).read() == 'test'

    time.sleep(2)

    assert view(
        'test_cache', {'test': 'macho'}
    ).cache_for(5, 'seconds').rendered_template == 'macho'

    assert open(glob.glob('bootstrap/cache/test_cache*')[0]).read() == 'macho'

    time.sleep(2)

    assert view(
        'test_cache', {'test': 'test1'}
    ).cache_for(1, 'second').rendered_template == 'macho'

    assert open(glob.glob('bootstrap/cache/test_cache*')[0]).read() == 'macho'
```

if you get that to work then it should be good to go. `import glob` too. 

The reason is because for some reason it keeps caching the old template over and over again so when I run the test and it works for 1 second, then the cache is expired after the test is ran but if I change the value to macho then it deletes the cache but it puts `test` in the cache instead of `macho`. That test case there should mimic what I'm doing.

that bottom test looks weird because although we are passing in `test1`, we should still get macho back because macho is still in the cache

masonite/view.py
```diff
@@ -51,3 +75,121 @@ def extend(self):
...
+        return False
+
+    def __is_expired_cache(self):
+        """
+        Check if cache is expired
+        """
+
+        # If is forever
+        if self.cache_time is None and self.cache_time is None:
+            return False
```
josephmancuso: I think the problem here is that this is returning `False` meaning that the cache is not expired. If it is forever then it should be `True` because it should act as if the cache is expired. So making this return True could fix your bug

masonite/view.py
```diff
@@ -127,15 +127,16 @@ def __is_expired_cache(self):
...
         # By seconds
         cache_type = self.cache_type.lower()
         calc = 0
         if cache_type == "second" or cache_type == "seconds":
-            calc = 1
+            # Set time now for
+            calc = self.cache_time
         elif cache_type == "minutes" or cache_type == "minute":
             calc = 60
         elif cache_type == "hours" or cache_type == 'hour':
```
josephmancuso: Let's also change this to `or hour` and not `or cache_type == 'hour'` to make it look cleaner

masonite/view.py
```diff
@@ -127,15 +127,16 @@ def __is_expired_cache(self):
...
+        if self.cache_time is None or self.cache_time is None and self.cache:
+            return True
 
         # By seconds
         cache_type = self.cache_type.lower()
         calc = 0
         if cache_type == "second" or cache_type == "seconds":
-            calc = 1
+            # Set time now for
+            calc = self.cache_time
```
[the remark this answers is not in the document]
mapeveri: It's true, I change it to 1.

tests/test_view.py
```diff
@@ -166,6 +169,24 @@ def test_view_cache():
...
+
+    assert open(glob.glob('bootstrap/cache/test_cache:*')[0]).read() == 'macho'
+
+    assert view(
+        'test_cache', {'test': 'macho'}
+    ).cache_for(1, 'second').rendered_template == 'macho'
+
+    time.sleep(1)
+
+    assert open(glob.glob('bootstrap/cache/test_cache:*')[0]).read() == 'macho'
```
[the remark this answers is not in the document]
mapeveri: Yes!

masonite/view.py
mapeveri: Ignore this!

masonite/drivers/CacheDiskDriver.py
```diff
@@ -0,0 +1,21 @@
+import os
+
+
+class CacheDiskDriver():
```
josephmancuso: Let's focus on expanding this cache driver now. all caching should have these methods:

* `store` -- stores the value inside the key
* `get` -- gets the value from the key
* `delete` -- deletes the value

Once we implement those methods which should have signatures like:

```
def store(self, key, value):

def get(self, key)

def delete(self, key)
```

Once those methods are created, test them to make sure they all work. So basically do the same type of assertions as the views (run the method and then `glob()` the file location to check the value is saved).

The idea here is that a lot of the logic of a view shouldn't be caching. it should be rendering views. So once we have tests for `Cache` then we can implement the `Cache` inside the view.

The cache should be used individually and not just for caching templates. We are just also going to use then to cache templates by passing them the `key` (the template name) and the `value` (the `self.rendered_template` value).

Again we should be able to use this completely outside caching templates like in a controller:

```python
def show(self, Cache):
    Cache.get('key') # returns the value
    Cache.store('key', 'value', '.txt') # stores the value inside `bootstrap/cache/key:888r8975445.txt`. 
```

Maybe pass in an extension here as the third parameter? it should default to `.txt` but inside the `View` class we can just pass `.html` 

You could either build this from scratch (should be simple) or you can look at a python library for this

masonite/providers/RouteProvider.py
```diff
@@ -93,7 +93,13 @@ def boot(self, WebRoutes, Route, Request, Environ):
                 # Get the data from the route. This data is typically the output
                 #     of the controller method
                 if not request.redirect_url:
```
josephmancuso: Ok so sometimes I want the view to return a string so it won't always have a `.rendered_template` attribute. We can do this simply by importing `from masonite.view import View` at the top and right here you'll need to do:

```python
# Get the data from the route. This data is typically the output
#     of the controller method
if not request.redirect_url:
    response = self.app.resolve(route.output)

    if isinstance(response, View):
        response = response.rendered_template


    self.app.bind(
        'Response',
        router.get(
            route.route,
            response
         )
    )
```

this will allow us to do something like:

```python
def show(self):
    return 'this'
```

which we were able to do before but now we are not since we changed this route provider.

masonite/view.py
```diff
@@ -24,15 +32,24 @@ def __init__(self):
     def render(self, template, dictionary={}):
```
josephmancuso: Add a docstring here

masonite/view.py
```diff
@@ -51,3 +68,76 @@ def extend(self):
     def share(self, dictionary):
         self.dictionary.update(dictionary)
         return self
+
+    def cache_for(self, time=None, type=None):
+        """
+        Set time and type for cache
+        """
+        self.cache = True
```
josephmancuso: space after the docstring here

masonite/view.py
```diff
@@ -51,3 +68,76 @@ def extend(self):
     def share(self, dictionary):
```
josephmancuso: add a docstring here. This should read `""" Updates the dictionary """`

masonite/view.py
```diff
@@ -24,15 +32,24 @@ def __init__(self):
...
-        return self.env.get_template(
-            template + '.html'
-        ).render(self.dictionary)
+        filename = template + '.html'
+        self.rendered_template = self.env.get_template(filename).render(
+            self.dictionary)
+
+        return self
 
     def composer(self, composer_name, dictionary):
```
josephmancuso: add a docstring here. This should read `""" Updates composer dictionary """`

masonite/view.py
```diff
@@ -51,3 +68,76 @@ def extend(self):
...
+        driver_cache = self.container.make('Cache')
+
+        # True is expired
+        return not driver_cache.is_valid(self.template)
+
+    def __get_cached_template(self):
+        """
+        Return the rendered template
+        """
+        driver_cache = self.container.make('Cache')
```
josephmancuso: space after this docstring

tests/test_view.py
```diff
@@ -105,3 +149,50 @@ def test_view_share_updates_dictionary_not_overwrite():
...
+
+    assert open(glob.glob('bootstrap/cache/test_cache:*')[0]).read() == 'test'
+
+    time.sleep(2)
+
+    assert view(
+        'test_cache', {'test': 'macho'}
+    ).cache_for(5, 'seconds').rendered_template == 'macho'
+
+    time.sleep(5)
```
josephmancuso: Change this to `time.sleep(2)` to speed up tests. it should still work and make tests faster

masonite/drivers/CacheDiskDriver.py
```diff
@@ -0,0 +1,70 @@
+import glob
+import os
+import time
+
+
+class CacheDiskDriver():
```
josephmancuso: Ok since this is a new class we'll need some tests for it. Just make some basic tests that test all the methods here and put it inside `test_cache.py`

masonite/drivers/CacheDiskDriver.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+
+        cache_path = self.config.DRIVERS['disk']['location'] + "/"
+        for template in glob.glob(cache_path + key + ':*'):
+            os.remove(template)
+
+    def cache_exists(self, key):
+        """
+        Check if the cache exists
+        """
```
josephmancuso: space after this docstring

masonite/drivers/CacheDiskDriver.py
```diff
@@ -0,0 +1,70 @@
...
+        path = self.config.DRIVERS['disk']['location'] + "/"
+        find_template = glob.glob(path + "/" + key + ":*")
+        if find_template:
+            return True
+        return False
+
+    def is_valid(self, key):
+        """
+        Check if a valid cache
+        """
```
josephmancuso: space after this docstring

masonite/view.py
```diff
@@ -51,3 +68,76 @@ def extend(self):
...
+        return self
+
+    def __create_cache_template(self, template):
+        """
+        Save in the cache the template
+        """
+
+        cache_type = self.cache_type.lower()
+        calc = 0
+        if cache_type == "second" or cache_type == "seconds":
```
josephmancuso: Try out something here. This looks really ugly and I'm trying to figure out how to make this cleaner. Possibly change:

```python
if cache_type == "second" or cache_type == "seconds":
```

to

```python
if cache_type in ("second", "seconds"):
```

That should make it look less ugly, shorter, and more readable

masonite/view.py
```diff
@@ -51,3 +68,76 @@ def extend(self):
...
+            calc = 60 * 60
+        elif cache_type == "days" or cache_type == 'day':
+            calc = 60 * 60 * 60
+        elif cache_type == "months" or cache_type == 'month':
+            calc = 60 * 60 * 60 * 60
+        elif cache_type == "years" or cache_type == 'year':
+            calc = 60 * 60 * 60 * 60 * 60
+        else:
+            # If is forever
+            return True
```
josephmancuso: This shouldn't return True, it should be `raise ValueError('{0} is not a valid caching type.'.format(cache_type))`

josephmancuso:
this is great! I'll take this, I just need a few minor changes I can probably do in 5 minutes so I'll just go ahead and do it. Great contribution! I took a while but it looks great!

mapeveri:
Great! Any problem, we keep seeing.

josephmancuso:
yes I'll do some more extensive testing and any bugs we can just create a test around it and work until its fixed

mapeveri:
Great, we can create new issues ;)

Issue #27, opened 2018-02-28
Add CSRF protection


josephmancuso:
So this will require a few moving parts here. 

So how CSRF works is that there is a random key generated (like `8yr8738b8b87b8br48gf84b7bv`) at the start of each request. This changes on every request so it needs to be random.

This token can be used inside forms so there needs to be a `{{ csrf_field() }}` function which just points to the `request.get_csrf()` method or something. Check the `HelpersProvider` for how helper functions work. Especially in templates (it uses the `View.share`) method.

Then when the request is submitted via POST, the CSRF token outputted by the `csrf_field()` function is checked against the one that was at the start of the request. If it's the same then the request was made by the current. If it wasn't then there is something wrong and it should throw an exception like `InvalidCSRFToken`

Now there is something to be aware of and that is that not all routes should have to be CSRF protected. It should be up to the developer to choose the ones that are not protected (all routes should be protected by default).

What I'm thinking of is that only new CSRF tokens should be generated if it is a GET request. This allows you to have a POST request and check if the CSRF token is valid

Please let me know if you have any questions

mapeveri:
I think it's pretty clear, what I think is that all routes must have csrf, now if a route does not need that protection, you should use some method or function, csrf_except or something similar.

josephmancuso:
Should we do it in the routes.py file?

```python
Get().route('url', 'controller').name('url').no_csrf()
```

or something like that?

mapeveri:
Yes, exactly.

josephmancuso:
or option 2 would be to create a list of routes in a configuration file but meh. I like the speed involved in  it

josephmancuso:
ok good

josephmancuso:
I say a good route here is that unless I'm thinking wrong, Get() should not have a `no_csrf()` method on it. So we can do like option 1 from the contracts issue where the Base class (check the masonite/routes.py) file should have a NotImplemented and only the Post class should override it and the Get will throw an exception. Just in case people don't realize it's only for POST requests.

CSRF is not needed for GET requests correct?

josephmancuso:
also remember that the `Request` object is instantiated before the server starts. So every time the request is made, its the same request object as the last request. Use that to your advantage

mapeveri:
It's true for GET requests not is necessary the csrf protection, is more, for GET request It doesn't add protection, but the opposite, makes a possible attack in the site.

PR #38, opened 2018-03-04
Add CSRF protection #27

mapeveri:


masonite/request.py
```diff
@@ -36,12 +39,33 @@ def __init__(self, environ=None):
 
         self.encryption_key = False
         self.container = None
+        # Generate csrf token
+        self.generate_csrf_token()
+
+    def generate_csrf_token(self):
```
josephmancuso: Ok so let's move all these csrf methods out of the request class and into a new `Csrf` class.

masonite/providers/HelpersProvider.py
```diff
@@ -22,3 +22,9 @@ def boot(self, View, ViewClass, Request):
         builtins.resolve = self.app.resolve
 
         ViewClass.share({'request': Request.helper, 'auth': Request.user})
+
+        # Share token csrf
+        token = Request.get_cookie('csrftoken')
+        ViewClass.share(
+            "<input type='hidden' name='csrftoken' value='" + token + "' />"
```
josephmancuso: This needs to be inside it's own `CSRFProvider` file. This should also be:

```python
ViewClass.share('csrf_field', "<input type='hidden' name='csrf_token' value='{0}' />".format(token))
```

then inside the template, we can do:

```html

<form action="" method="">
    {{ csrf_field }}
    <input type=.....
</form>

```

masonite/middlewares/VerifyCsrfToken.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Verify csrfk token middleware
+    """
+
+    exempt = []
+
+    def __init__(self, Request):
+        self.request = Request
+
+    def befofe(self):
```
josephmancuso: make sure the typo in `befofe` is `before`

masonite/middlewares/VerifyCsrfToken.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class VerifyCsrfToken(object):
+    """
+    Verify csrfk token middleware
+    """
+
+    exempt = []
+
+    def __init__(self, Request):
```
josephmancuso: When you add the `CSRFProvider` then you'll need to also but `CSRF` in this constructor

masonite/middlewares/VerifyCsrfToken.py
```diff
@@ -0,0 +1,33 @@
...
+        self.request = Request
+
+    def befofe(self):
+        request = self.request
+        if request.is_post():
+            token = request.input('csrftoken')
+            if not request.verify_csrf_token(token) and not self.__in_except():
```
josephmancuso: When you move these verify methods into it's own `CSRF` class make sure you use that here.

masonite/providers/CsrfProvider.py
```diff
@@ -0,0 +1,19 @@
...
+from masonite.auth.Csrf import Csrf
+
+
+class CsrfProvider(ServiceProvider):
+
+    wsgi = True
+
+    def register(self):
+        request = self.app.make('Request')
+        self.app.bind('Request', request)
```
josephmancuso: remove the `self.app.bind('Request', request)`

the request is already binded to the container in the `AppProvider`

You're overwriting it here

masonite/middlewares/CsrfMiddleware.py
```diff
@@ -0,0 +1,36 @@
...
+
+    exempt = []
+
+    def __init__(self, Request, CSRF):
+        self.request = Request
+        self.csrf = CSRF
+
+    def before(self):
+        if self.request.is_post():
+            token = self.request.input('csrftoken')
```
josephmancuso: Notice that you're getting the `.input('csrftoken')` but inside the input on line 19 of `CsrfProvider` it is `csrf_token`. I'd go with the underscore one

PR #48, opened 2018-03-21
New helpers for routes #47

### LINKED ISSUE #47: New helpers for routes
For example:

```
from masonite.routes import Get
def get(url, controller)
    return Get().route(url, controller)
```

Add this helpers for GET, POST, PUT, and DELETE.

mapeveri:
Closes #47

### DIFF
--- a/tests/test_helpers_urls.py
+++ b/tests/test_helpers_urls.py
@@ -1,0 +1,21 @@
+from masonite.helpers import urls_helpers
+
+
+def test_get_sets_route():
+    assert urls_helpers.get('test', None)
+
+
+def test_post_sets_route():
+    assert urls_helpers.post('test', None)
+
+
+def test_put_sets_route():
+    assert urls_helpers.put('test', None)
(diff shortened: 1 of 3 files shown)


tests/test_helpers_urls.py
```diff
@@ -0,0 +1,21 @@
+from masonite.helpers import urls_helpers
+
+
+def test_get_sets_route():
+    assert urls_helpers.get('test', None)
```
josephmancuso: In addition to all these asserts assert that they are instances of the correct HTTP classes. For instance something like:

```python
from masonite.routes import Get
assert isinstance(urls_helpers.get('test', None, Get)
```

that might not actually work right there because it might not be an actual instance. I forget how Python checks for instances but you get the point. Let me know if you have any questions about that. Do that for all the tests.

mapeveri: @josephmancuso Check now please :)

Issue #52, opened 2018-03-30
Add Sessions


josephmancuso:
Since the request is technically a singleton throughout the life of the server (it is instantiated before the server is started) we can technically have sessions. We would need to create a `session()` method on the request class which will generate a `_session` attribute on the request class that might look something like:

```python

_session = {
    '28.029.291.199': {
        'username': 'Joseph'
    }
}
```

Then we can use something like:

```python
# sets the session variable
Request.session('username', 'Mike')
```

or just use:

```python
# sets the session variable
Request.session('username')
# Returns Mike
```

I have to make sure that somehow this is secure. Using an IP address might actually be insecure because it might be able to be spoofed and now we have cross session access. I'll need a way to somehow ensure that WSGI doesn't allow a way to set the REMOTE_ADDR as an environ variable (I think that even if one is sent it will be converted to HTTP_REMOTE_ADDR) so it should be fine. There is something called `X-HTTP-FORWARDED-FOR`. Might need to check that.

https://stackoverflow.com/questions/7835030/obtaining-client-ip-address-from-a-wsgi-app-using-eventlet?lq=1

This should also be it's own class and the request class should inherit it. Call the class `Session` which should not be able to be instantiated directly (and abstract class). Put this class inside a new file inside the base of the project called `session.py`. So this should be inside `masonite/session.py` and put the class Session inside of that.

Then import it into the request.py and inherit the Request class from it.

@mapeveri Is this something you'd want to do? It's pretty straight forward.

mapeveri:
@josephmancuso Yes, I think a time to do it. One thing, I would like the methods to be 'set' and 'get' to set and get the session. What do you think?

josephmancuso:
so a `set_session` and `get_session`. Although it "looks" like it's more explicit and easier to understand what's happening, I'm not entirely sure that adds any value over the normal way.

I'm not a huge fan of `get_` and `set_` the two methods. I like it when it's not going to be primarily developer facing. For example if it's just an internal method that we use for `core` only. If it's going to be front facing I'd like it to be as clean as possible.

If you'd like to know how to implement the one parameter get and 2 parameter set then look at the new `header()` method on the Request class.

josephmancuso:
It's kind of a random thought I had when making the `header` method (the 1 method 2 parameter thing). Maybe we'll change it for Masonite 1.6 but let's go with that for now. It is very opinionated I know lol but we're taking a gamble on it :)

mapeveri:
[...]

PR #55, opened 2018-04-02
Add Sessions #52

mapeveri:


masonite/session.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def get(self, key):
+        """
+        Get a session from object _session
+        """
+        ip = self.__get_client_address()
+        try:
+            session = self._session[ip][key]
+        except KeyError:
+            session = ''
```
josephmancuso: Let's make this None instead. It makes sense that if you try to get something and it doesn't exist then it should return None. If it returns an empty string that seems odd to me

I think this should actually just look like:

```python
ip = self.__get_client_address()

try:
    return self._session[ip][key]
except KeyError:
    return None
```

just to shorten it up. but do some testing on that, that might not work. Doing that should make the same tests pass but you just have to check for `None` instead of a blank string `''`

masonite/session.py
```diff
@@ -0,0 +1,47 @@
...
+
+        return session
+
+    def set(self, key, value):
+        """
+        Set a new session in object _session
+        """
+        ip = self.__get_client_address()
+        self._session[ip] = {key: value}
```
josephmancuso: so if you see here, we can only set 1 value to the session. every time you set something it can only contain 1 dictionary in the IP address. 

This test should pass:

```python
def test_session_request():
    REQUEST.session.set('username', 'pep')
    REQUEST.session.set('password', 'secret')
    assert REQUEST.session.get('username') == 'pep'
    assert REQUEST.session.get('password') == 'secret'
```

but it fails because it overwrites the previous set.

This piece of code might have to look something like:

```python
self._session[ip]['key'] = value
```

Now every time you set a different key, it will just keep adding key and value pairs to the session or it will overwrite it.

if that doesn't work you might have to just update the dictionary instead so something like:

```python
self._session[ip].update({key: value})
```

Something like that. Just play with that. Either way that test above should pass. I should be able to set multiple keys on the session

masonite/session.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def reset(self):
```
josephmancuso: This should be good. Um lets also also do 2 more things here. We need a way to:

1. Reset all session data from a certain IP address.

So you can just specify a second parameter:

```python
def reset(self, string ip=None, bool current_user=False):
```

then just do those actions so if the delete from the the ip key if one is passed or delete from the current user (using your private method of getting the IP address) if that is True

The `string` and `bool` prefixes are just for you so you get an idea of what should be passed in.

tests/test_session.py
```diff
@@ -0,0 +1,43 @@
...
+    'SERVER_PORT': '8000',
+    'PATH_INFO': '/',
+    'SCRIPT_NAME': ''
+}
+
+REQUEST = Request(wsgi_request).key(
+    'NCTpkICMlTXie5te9nJniMj9aVbPM6lsjeq5iDZ0dqY=')
+
+
+def test_session_request():
```
josephmancuso: again just get this test to pass:

```python
def test_session_request():
    REQUEST.session.set('username', 'pep')
    REQUEST.session.set('password', 'secret')
    assert REQUEST.session.get('username') == 'pep'
    assert REQUEST.session.get('password') == 'secret'
```

tests/test_session.py
```diff
@@ -0,0 +1,43 @@
...
+    REQUEST.session.set('username', 'pep')
+    assert REQUEST.session.get('username') == 'pep'
+
+
+def test_reset_session():
+    REQUEST.session.set('username', 'pep')
+    REQUEST.session.reset()
+    assert REQUEST.session.get('username') == ''
```
josephmancuso: this should be:

```python
assert REQUEST.session.get('username') == None
```

if you want you can also do:

```python
assert REQUEST.session.get('username') is None
```

PR #60, opened 2018-04-08
Cookie expiration and Sessions

### LINKED ISSUE #59: change cookies to set an expire date
right now all cookies are set without an expiration date which makes them session cookies essentially. If a user closes out there browser completely, the cookie will be deleted. Make all cookies be defaulted to forever but can be changed by using the `time` parameter.

Something like:

```python

Request.cookie('key', 'value', time='1 minute')

```

josephmancuso:
- [x] Needs documentation
- [x] Needs Review

Closes #59, #63 

# Running Changelog

* Added sessions with memory and cookie drivers
* added cookie expirations dates
* Added a way to delete cookies
* added coveralls
* improved test coverage


mapeveri:
The PR is ok, the tests would be missing?

josephmancuso:
Yes you're right. I added a test now.

PR #61, opened 2018-04-08
Fix with run tests and improvements pep8

mapeveri:


masonite/session.py
```diff
@@ -36,44 +34,39 @@ def set(self, key, value):
...
 
-
     def flash(self, key, value):
         """
         Add temporary data to the session
         """
 
         ip = self.__get_client_address()
-        if not ip in self._flash:
+        if not (ip in self._flash):
```
josephmancuso: Is this really PEP 8? It looks like a useless case of parenthesis. 

(ip in self._flash) is setting a boolean value (True or False) and you are "not"ing it.

maybe we should change it to `if ip not in self._flash` to make it more "Pythonic" (I hate that word)

requirements.txt
```diff
@@ -10,4 +10,5 @@ requests==2.18.4
 boto3==1.5.24
 tldextract==2.2.0
 pusher==1.7.4
-ably==1.0.1
\ No newline at end of file
+ably==1.0.1
+pendulum==1.4.4
```
josephmancuso: good catch here 👍

setup.py
```diff
@@ -35,7 +35,8 @@
         'cryptography==2.1.4',
         'PyMySQL==0.8.0',
         'psycopg2==2.7.3.2',
-        'masonite-entry>=0.0.0,<=0.9.99'
+        'masonite-entry>=0.0.0,<=0.9.99',
+        'pendulum==1.4.4',
     ],
     description='The core for the Masonite',
```
josephmancuso: Can we actually change this to `description='The core for the Masonite framework',`

masonite/session.py
mapeveri: Yes, if ip not in self._flash it's better :)

setup.py
mapeveri: Perfect

mapeveri:
Ready :)

Issue #53, opened 2018-03-30
Create drivers for sessions


mapeveri:
This issue depends on #52. What is needed is to add drivers to store the sessions, for example Redis, etc.

mapeveri:
I will see about making this issue.

josephmancuso:
Great this should be fairly simple. Just take all of the logic of the session and make it a `memory` driver. Then you'll have to make a `config/session.py` file where we can put our driver configuration files

mapeveri:
Please gitter ;)

PR #63, opened 2018-04-10
Create drivers for sessions #53

mapeveri:


masonite/drivers/SessionCookieDriver.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+
+        cookies = []
+        if 'HTTP_COOKIE' in self.environ and self.environ['HTTP_COOKIE']:
+            cookies_original = self.environ['HTTP_COOKIE'].split(';')
+            for cookie in cookies_original:
+                if cookie.startswith('s_'):
+                    data = cookie.split("=")
+                    result = {'key': data[0], 'value': data[1]}
+                    cookies.append(result)
```
josephmancuso: this is a dictionary so I believe it should be update and not append

masonite/drivers/SessionCookieDriver.py
```diff
@@ -0,0 +1,86 @@
...
+        Collect data from session and flash data
+        """
+
+        cookies = []
+        if 'HTTP_COOKIE' in self.environ and self.environ['HTTP_COOKIE']:
+            cookies_original = self.environ['HTTP_COOKIE'].split(';')
+            for cookie in cookies_original:
+                if cookie.startswith('s_'):
+                    data = cookie.split("=")
+                    result = {'key': data[0], 'value': data[1]}
```
josephmancuso: this shouldn't be `result = {'key': data[0], 'value': data[1]}` it should be 

`result = {data[0],  data[1]}`

Issue #68, opened 2018-04-17
Add a flag for HTTP_ONLY cookies


josephmancuso:
Right now all cookies are set with an HTTP only flag. This prevents javascript from reading cookies. This was the intended result because it was a security measure to prevent malicious javascript code from reading cookies (like retrieving sessions)

Sometimes you may wish to actually set cookies and have your javascript code use it so there should be a parameter to set it:

```python
request().cookie('key', 'value', encrypt=False, http_only=False)
```

PR #70, opened 2018-04-19
Add a flag for HTTP_ONLY cookies #68

mapeveri:


masonite/request.py
```diff
@@ -194,8 +195,13 @@ def cookie(self, key, value, encrypt=True, path='/', expires=''):
         if expires:
             expires = "Expires={0};".format(cookie_expire_time(expires))
 
+        http_only_text = ""
+        if http_only:
+            http_only_text = "HttpOnly"
+
```
josephmancuso: hmm I'm not really a big fan of the random `http_only_text` ...

can we do something like:

```python
def cookie(..., http_only='HttpOnly')

if not http_only:
    http_only = ''
...
```

and then the rest looks good.

The idea here is that we developers will never need to specify it as True because it's always true. for example they will never do:

```python
request.cookie('key', 'value', http_only=True)
```

since it's always true by default. So with that we can just specify the string and only check for a falsy value (`False`)

since we set it to an empty string it obviously won't show up when setting the cookie

tests/test_cookie_signing.py
```diff
@@ -62,3 +62,8 @@ def test_set_and_get_cookie_with_existing_cookie():
     REQUEST.environ['HTTP_COOKIE'] = 'cookie=true'
     REQUEST.cookie('test', 'testvalue')
     assert REQUEST.get_cookie('test') == 'testvalue'
+
+
+def test_set_and_get_cookie_http_only():
+    REQUEST.cookie('test', 'testvalue', http_only=True)
+    assert REQUEST.get_cookie('test') == 'testvalue'
```
josephmancuso: let's add an additional test here.

again the `http_only=True` doesn't really make sense. there should be a test like:

```python
REQUEST.cookie('test', 'testvalue', http_only=True)
assert REQUEST.get_cookie('test') == 'testvalue'

assert 'HttpOnly' in REQUEST.cookies[0][1] # [0] gets the first tuple and then the [1] gets the tuples value (where the cookie information is)
```

so just get rid of the http_only=True and it should be http_only=False and then check that the `HttpOnly` doesn't exist

josephmancuso:
The reduction in code coverage for the driver stuff is fine. it just means that travis hasn't triggered the environment variables for them. slightly annoying.

mapeveri:
@josephmancuso check now please :)

tests/test_cookie_signing.py
```diff
@@ -64,6 +64,13 @@ def test_set_and_get_cookie_with_existing_cookie():
...
 
-def test_set_and_get_cookie_http_only():
-    REQUEST.cookie('test', 'testvalue', http_only=True)
+def test_set_and_get_cookie_with_http_only():
+    REQUEST.cookie('test', 'testvalue')
+    assert REQUEST.get_cookie('test') == 'testvalue'
+    assert 'HttpOnly' in REQUEST.cookies[0][1]
+
+
+def test_set_and_get_cookie_without_http_only():
```
josephmancuso: Ok so the above test is good but the test below is bad. The test below should be something like:

```python
REQUEST.cookies = [] # reset the cookie list
REQUEST.cookie('test', 'testvalue', http_only=False)
assert 'test=testvalue; Path=/' in REQUEST.cookies[0][1]
```

I'm not sure the test below shouldn't have passed because it implies that the string is empty.

So just for your information:

REQUEST.cookies[0] should return `('Set-Cookie', 'key=value; HttpOnly ... blah blah')`

then REQUEST.cookies[0][1] should return `'key=value; HttpOnly ... blah blah'`

make sure the cookie is constructed correctly

mapeveri:
@josephmancuso There I made ammend. In my case, the tests pass.

One question, If I pass httpOnly = 'test' I would not have to validate that it is correct?

tests/test_cookie_signing.py
josephmancuso: Ok so just change this test here to this:

```python
def test_set_and_get_cookie_without_http_only():
    REQUEST.cookies = []
    REQUEST.cookie('test', 'testvalue', http_only=False, encrypt=False)
    assert REQUEST.get_cookie('test', decrypt=False) == 'testvalue'
    assert 'test=testvalue; Path=/' in REQUEST.cookies[0][1]
```

mapeveri: Ready, added the these test and request class in the method cookie changed the declaration like this: 

def cookie(self, key, value, encrypt=True,
               http_only="HttpOnly;", path='/', expires=''):

Because otherwise the test failed.

josephmancuso:
Awesome. Looks great! when I get some time I'll QA test it before merging it in. Probably tomorrow

mapeveri:
@josephmancuso great!. Thanks!

tests/test_cookie_signing.py
```diff
@@ -62,3 +62,16 @@ def test_set_and_get_cookie_with_existing_cookie():
     REQUEST.environ['HTTP_COOKIE'] = 'cookie=true'
     REQUEST.cookie('test', 'testvalue')
     assert REQUEST.get_cookie('test') == 'testvalue'
+
+
+def test_set_and_get_cookie_with_http_only():
```
josephmancuso: Instead make the test:

```python
def test_set_and_get_cookie_with_http_only():
    REQUEST.cookies = []
    REQUEST.cookie('test', 'testvalue', encrypt=False)
    assert REQUEST.get_cookie('test', decrypt=False) == 'testvalue'
    assert 'HttpOnly' in REQUEST.cookies[0][1]
    assert 'test=testvalue; HttpOnly; Path=/' in REQUEST.cookies[0][1]
```

mapeveri: Check please :)

josephmancuso:
Perfect. I'll merge after the tests run

PR #75, opened 2018-04-23
added support for application/json type requests

josephmancuso:
This PR will add application/json support. The current way Masonite gets input data (post get and post using cgi.FieldStorage) currently does not support application/json (for some dumb reason) so if the server receives an application/form-data it will be fine but if it receives an application/json it will not be able to get it.

Now masonite will be able to handle this and stores the json receives in a payload input:

```python
request().input('payload') # returns a dictionary of the JSON request sent to the server
```


josephmancuso:
I have an idea to bump the coverage. I'll need to make the `wsgi.input` an object that has a read class that returns a json string I think.

PR #85, opened 2018-04-25
Improve container

josephmancuso:


josephmancuso:
We actually cut down A LOT here in the logic of the container. I realized it was extremely more complex than it needed to be and also took longer than it needed to. This should also speed up the framework a bit

PR #89, opened 2018-04-26
added global views

josephmancuso:
This adds a new way for global templates. The use case here is when we want to either use templates that are outside of our application such as third party views or even just views that are not inside of `resources/templates`:

Let's say we do:

    $ pip install masonite-billing

Then we might want some view like a unsubscribe view or something:

normally we have something like:

```python
def show(self):
    return view('unsubscribe')
```

which looks inside our `resources/templates` directory. Well now we can look globally for them:

```python
def show(self):
    return view('/billing/views/unsubscribe')
```

which will now look in the billing module we just installed, and inside the views directory for a unsubscribe.html view which we can now return.

Not only is this good for rendering third party packages but it can also even be useful for entire features. For example Django has that dashboard. Well now because controllers and views can be accessible globally, we could do something like:

    $ pip install masonite-dashboard

and inside `routes/web.py` we can do:

```python
from dashboard.routes import DashboardRoutes

# DashboardRoutes is a list that contains something like:
# [
#     Get().route('/dashboard', '/dashboard.controllers.DashboardController@show'),
#     Get().route('/dashboard/@model', '/dashboard.controllers.ModelController@show'),
#     Get().route('/dashboard/@model/edit', '/dashboard.controllers.ModelController@update'),
#     Post().route('/dashboard/@model', '/dashboard.controllers.ModelController@create'),
#      .... etc etc ....
# ]

...

ROUTES += DashboardRoutes
```

which will literally just take over a route (in this case the route is `/dashboard`)

@mapeveri Thoughts? how cool is this?


mapeveri:
It's great. I like it. I do not have any suggestions, I think it's perfect.

Issue #93, opened 2018-04-29
Create route groups


josephmancuso:
Thought of something really cool. We can have route groups. Something like:

```python
from masonite.helpers.routes import group, get, post

ROUTES = [
    get('/some/url', 'Controller@show'),
    ....
    group('/dashboard', [
        get('/user', 'UserController@show'),  # returns '/dashboard/user'
        post('/user', 'UserController@store'),
        ...
    ])
]
```

The problem here is that this will have to be apart of the list. group will need to return a list but then you will end up with something like:

```python
[
    get(..),
    post(..)
    [
        get(..),
        post(..)
    ]
]
```

This obviously won't work because we need a single list. So what we could do is something like:

```python
from masonite.helpers.routes import group, get, post

ROUTES = [
    get('/some/url', 'Controller@show'),
    ....
    *group(prefix='/dashboard', [
        get('/user', 'UserController@show'), # returns '/dashboard/user'
        post('/user', 'UserController@store'),
        ...
    ])
]
```

(notice the star there which should unpack the list) But I think that's ugly and people are going to forget that

or can `group` could possibly return an unpacked list?

```python
def group(url, route_list):
    return *route_list # will this work? I think so
```

the group function will need something like:

```python
def group(prefix, route_list):
    # prepend the prefix ('/dashboard')
    for route in route_list:
        route.route_url = prefix + route.route_url # now '/dashboard/user'

    return *route_list
```

@mapeveri What do you think about this?

This can actually ship with Masonite 1.6.1

mapeveri:
A user level, for me, I like this example:

```
from masonite.helpers.routes import group, get, post

ROUTES = [
    get('/some/url', 'Controller@show'),
    ....
    group('/dashboard', [
        get('/user', 'UserController@show'),  # returns '/dashboard/user'
        post('/user', 'UserController@store'),
        ...
    ])
]
```
This is very clear. 
Maybe in the group of functions, join the routes and return only one.?

josephmancuso:
> Maybe in the group of functions, join the routes and return only one.

can you elaborate on that? I'm playing around with this feature now and apparently you cant `return *list` ..

mapeveri:
Sure, when the list are like this:

```
[
    get(..),
    post(..)
    [
        get(..),
        post(..)
    ]
]
```

The function must return something like that:

```
[
    get(..),
    post(..)
    get(/dashboard/users...),
    post(/dashboard/users...)
]
```

[...]

PR #94, opened 2018-04-29
Create route groups #93

mapeveri:


mapeveri:
Ready :)

masonite/helpers/routes.py
```diff
@@ -34,3 +34,16 @@ def patch(url, controller):
     Shortcut for PATCH http
     """
     return Patch().route(url, controller)
+
+
+def group(url, route_list):
+    """
+    Group route
+    """
+    if url[-1:] != '/':
```
josephmancuso: what is this here? this looks really confusing. Are you checking if the url starts with a forward slash?

If so you should just do something like:

```python
if not url.startswith('/'):
    # do logic
```

but even still I don't think we should check for this. We don't check in any other routes. All routes should start with a slash anyway. If we start doing this check here then we need to perform this check on all routes. If we are going to make this change then I say we do it in some 2.x release instead.

routes/web.py
```diff
@@ -1,7 +1,12 @@
+from masonite.helpers.routes import group
 from masonite.routes import Get
 
 
 ROUTES = [
     Get().route('/test', None).middleware('auth'),
     Get().domain('test').route('/test', None).middleware('auth'),
+    group('/example', [
+        Get().route('test/1', 'TestController@show'),
```
josephmancuso: this route should here should have forward slashes in the beginning. let's keep it in line with the rest of the routes

tests/test_routes.py
```diff
@@ -76,5 +77,16 @@ def test_route_gets_controllers():
...
     assert Get().route('test/url', '/app.http.test_controllers.TestController@show')
 
+
 def test_route_doesnt_break_on_incorrect_controller():
     assert Get().route('test/url', 'BreakController@show')
+
+
+def test_group_route():
+    routes = group('/example', [
+        Get().route('test/1', 'TestController@show'),
```
josephmancuso: same thing here. Change this line to `Get().route('/test/1', 'TestController@show')` and get this same test to pass. Good test.

masonite/helpers/routes.py
mapeveri: No, I only check if it has "/" at the end. I erase it?

routes/web.py
mapeveri: Ok

tests/test_routes.py
mapeveri: Ok

masonite/helpers/routes.py
josephmancuso: oh ok ummm then that's fine actually. But then do `url.endswith('/')`

josephmancuso: actually you are appending a `/` here. Don't do that. Just get rid of this actually

mapeveri: Ok

mapeveri:
Check please :)

josephmancuso:
Looks awesome! I'm gonna make some minor variable name changes and merge this in

Issue #105, opened 2018-05-04
UploadManager should detect drivers bound to the container after the manager.


aisola:
This one might be a bit hard for me to explain, but I'll give it a shot.

In creating a service provider adding an [upload driver for Backblaze](https://github.com/aisola/masonite-backblaze), I noticed that installing the UploadManager doesn't seem to detect upload drivers bound to the container after the manager is bound _regardless of whether the new driver is bound in the same service provider or a new one_.

Steps to reproduce:
* Create a new masonite project and install per usual.
* Install [upload driver for Backblaze](https://github.com/aisola/masonite-backblaze).

```
git clone git@github.com:aisola/masonite-backblaze.git
cd masonite-backblaze
pip install .
```

* Alternatively to the above, you can create your own test driver and create a provider for it.
* Put provider in config/application.py after the 'Third Party' comment

If at this point, you check the container by launching a python console, importing wsgi, and listing the providers from the container, you will note that `UploadBackblazeDriver` _is_ in fact in the container.

* Change the `STORAGE_DRIVER` variable in your `.env` file to be `backblaze` (or your own test driver name).

At this point, running the craft command fails with the following traceback:

```

Traceback (most recent call last):
  File "/home/acisola/workspace/python3/snippets/venv/lib/python3.4/site-packages/masonite/managers/Manager.py", line 36, in create_driver
    '{0}{1}Driver'.format(self.driver_prefix, driver)
  File "/home/acisola/workspace/python3/snippets/venv/lib/python3.4/site-packages/masonite/app.py", line 32, in make
    raise MissingContainerBindingNotFound("{0} key was not found in the container".format(name))
masonite.exceptions.MissingContainerBindingNotFound: UploadBackblazeDriver key was not found in the container

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/craft", line 11, in <module>
    load_entry_point('masonite-cli==2.0.1', 'console_scripts', 'craft')()
  File "/usr/local/lib/python3.4/dist-packages/pkg_resources/__init__.py", line 480, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/lib/python3.4/dist-packages/pkg_resources/__init__.py", line 2691, in load_entry_point
    return ep.load()
  File "/usr/local/lib/python3.4/dist-packages/pkg_resources/__init__.py", line 2322, in load
    return self.resolve()
[...]

PR #106, opened 2018-05-04
Stop instantiating XManagers in ServiceProvider.register

aisola:
Fixes MasoniteFramework/core#105


masonite/providers/BroadcastProvider.py
```diff
@@ -14,7 +14,7 @@ def register(self):
         self.app.bind('BroadcastConfig', broadcast)
         self.app.bind('BroadcastPusherDriver', BroadcastPusherDriver)
         self.app.bind('BroadcastAblyDriver', BroadcastAblyDriver)
-        self.app.bind('BroadcastManager', BroadcastManager(self.app))
+        self.app.bind('BroadcastManager', BroadcastManager)
```
josephmancuso: So I think we should still instantiate the drivers the `self.app` but inside the contractor we can get rid of creating the driver. If that makes sense. Let me know if you have anymore questions.

josephmancuso:
btw, sometimes one of the cache tests fail. Don't worry about that. It's because one of the cache files aren't deleting correctly

josephmancuso:
I think the solution here is to keep everything like it was but not create the driver until inside the boot method. I'll do some testing

josephmancuso:
ok yeah after some testing it looks like all you need to do is remove the:

```python
if container:
    self.create_driver()
```

from the constructor inside the Manager class (`masonite/managers/Manager.py`)

aisola:
Updated with your new solution. That makes sense. I've also bumped the version number as you asked.

Sorry about my misunderstanding with the branches. I'll be sure to base it differently in the future.

josephmancuso:
The tests just keep failing because now we don't create drivers on instantiating. It looks like only the failing tests are in the `tests/test_managers_mail_manager.py` file. Tests are pretty straight forward in Masonite. Everything is function based for now, I'll probably move to test classes eventually. You can see failing tests here: https://travis-ci.org/MasoniteFramework/core/jobs/375008546

or run tests locally by doing : `python -m pytest`

aisola:
Ah shoot, sorry. Updating tests now.

aisola:
Looks like all tests pass now. https://travis-ci.org/MasoniteFramework/core/jobs/375019047

josephmancuso:
Everything looks great! it totally makes sense to not create a driver in the constructor. This way everything can be registered first before deciding on what to create. Merging and releasing.

josephmancuso:
@aisola also, if you want to hop in the Gitter chat: https://gitter.im/masonite-framework/Lobby

Issue #114, opened 2018-05-07
Request.redirectTo doesn't actually redirect the request


aisola:
`Request.redirectTo` doesn't actually redirect the request. Additionally, if the app developer doesn't return a real response after the call, it also causes an error (traceback below).

```
ERROR:waitress:Exception when serving /home
Traceback (most recent call last):
  File "/home/acisola/workspace/test_project/bootstrap/start.py", line 49, in app
    container.resolve(located_provider.boot)
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/masonite/app.py", line 65, in resolve
    return obj(*provider_list)
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/masonite/providers/StartResponseProvider.py", line 15, in boot
    data = bytes(Response, 'utf-8')
TypeError: encoding or errors without a string argument

During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/waitress/channel.py", line 338, in service
    task.service()
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/waitress/task.py", line 169, in service
    self.execute()
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/waitress/task.py", line 399, in execute
    app_iter = self.channel.server.application(env, start_response)
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/whitenoise/base.py", line 66, in __call__
    return self.application(environ, start_response)
  File "/home/acisola/workspace/test_project/bootstrap/start.py", line 51, in app
    container.make('ExceptionHandler').load_exception(e)
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/masonite/exception_handler.py", line 28, in load_exception
    self._render()
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/masonite/exception_handler.py", line 48, in _render
    'open': open
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/masonite/view.py", line 79, in render
    self.dictionary)
  File "/home/acisola/workspace/test_project/venv/lib/python3.4/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
[...]

Issue #113, opened 2018-05-07
Redirects in Middleware.before don't actually redirect the request


aisola:
Noticed this while making a middleware for my application.

Take the default authentication middleware provided by `craft auth`.

```
class AuthenticationMiddleware:
    ''' Middleware To Check If The User Is Logged In '''

    def __init__(self, Request):
        ''' Inject Any Dependencies From The Service Container '''
        self.request = Request

    def before(self):
        ''' Run This Middleware Before The Route Executes '''
        if not self.request.user():
            self.request.redirectTo('login')

    def after(self):
        ''' Run This Middleware After The Route Executes '''
        pass
```

I expect this middleware to redirect the request if the user is not logged in. As it turns out, the only reason why the `HomeController` redirects with no logged in user is because `HomeController.show` includes:

```
if not Auth(Request).user():                                                                               
            Request.redirect('/login')
```

josephmancuso:
Oh interesting. I'll take a long at this in a few

aisola:
I think that (for some reason) it's not setting `Request.redirect_url` because if I check that value in the controller it shows as `False`.

PR #115, opened 2018-05-08
Change around the redirect semantics.

aisola:
Fixes MasoniteFramework/core#114
Fixes MasoniteFramework/core#113

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         'masonite.contracts',
         'masonite.helpers',
     ],
-    version='1.6.4',
+    version='1.6.6',
     install_requires=[
         'validator.py==1.2.5',
         'cryptography==2.2.2',
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         'masonite.contracts',
         'masonite.helpers',
     ],
-    version='1.6.4',
+    version='1.6.5',
     install_requires=[
         'validator.py==1.2.5',
         'cryptography==2.2.2',
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
         'masonite.contracts',
         'masonite.helpers',
     ],
-    version='1.6.5',
+    version='1.6.4',
     install_requires=[
         'validator.py==1.2.5',
         'cryptography==2.2.2',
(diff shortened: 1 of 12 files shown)


josephmancuso:
This looks awesome. Let me pull it in and run some user tests on it. Hey what do you think of this:

1. Change the redirect_to back to redirectTo.
2. Merge this into 1.6 branch to fix the bug. 
3. Make a new PR and change it back to redirect_to
4. merge it into develop for the 2.0 branch?

In the mean time I'll test this out

josephmancuso:
I like it I just found 1 bug. 

in StartResponse there are these lines:

```python
            self.app.bind('Headers', [
                ('Location', Request.compile_route_to_url())
            ] + Request.get_cookies())
```

Since we made changes to the compile_route method this should I believe change to:

```python
            self.app.bind('Headers', [
                ('Location', Request.redirect_url) # <- Change here
            ] + Request.get_cookies())
```

josephmancuso:
Also, are we using the `request.redirect_route` attribute at all anymore?

aisola:
Oh good catch. I didn't think so, but I left it in there just in case. We aren't using it in any of the rewritten redirects. I forgot to mention it. Should it be removed?

josephmancuso:
we'll need to keep it for 1.6. Let's remove it in the PR for develop

aisola:
Cool fair enough.

aisola:
`StartResponseProvider` has now been fixed.

josephmancuso:
Hm it looks like you merged the develop branch with the 1.6 branch so this PR is also adding the changes to resolving controller constructors. If it too annoying to cherry pick the changes let me know and I'll just merge this into 1.6 anyway and move that from the 2.0 docs to the 1.6 docs.

josephmancuso:
You might actually just be able to `git revert <commit-hash-id>`

aisola:
Okay @josephmancuso, I believe that should fix it.

setup.py
```diff
@@ -15,7 +15,7 @@
         'masonite.contracts',
         'masonite.helpers',
     ],
-    version='1.6.5',
+    version='1.6.4',
```
josephmancuso: This should be 1.6.6 so we can release the next bug fix

aisola: Oh shoot! Sorry! I'll get that done.

aisola:
Bumped

josephmancuso:
🥇 🥇

PR #116, opened 2018-05-08
Add the tinker command

aisola:
Added this out of inspiration. I've based this off of develop so that we may have more discussion and push it out with the 2.0 release.

In essence, the `craft tinker` command is nothing more special than a python console with the container pre-loaded as a variable `app`. What is nice about it is that it allows us to more easily tinker around with the internals of the application either for testing or for seeding information before creating the store routes in our app. For instance you could do the following to retrieve all users.

```
>>> User = app.make('User')
>>> user = User.all()
>>> for user in users:
...     print(user.id)
...
1
2
3
4
```

For beginners, this will be a valuable tool for them to be able to play around with the container in an interactive way.

If there's anything else you want me to import by default into this console, that can easily be done.

### DIFF
--- a/masonite/commands/TinkerCommand.py
+++ b/masonite/commands/TinkerCommand.py
@@ -1,0 +1,29 @@
+import code
+import sys
+from cleo import Command
+
+BANNER = """Masonite Python {} Console
+This interactive console has the following things imported:
+    container as 'app'
+
+Type `exit()` to exit."""
+
+
+class TinkerCommand(Command):
+    """
+    Run a python shell with the container pre-loaded.
+
+    tinker
+    """
+
+    def handle(self):
+        from wsgi import container
+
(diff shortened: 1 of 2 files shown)


josephmancuso:
No way. @mapeveri suggested this a while ago but I couldn't figure out how to do this lol. I'm gonna review now

josephmancuso:
wow this is awesome.

josephmancuso:
are there other things we should load in?

aisola:
I'm not sure. I was thinking that it might be helpful to automatically load in the models, but I figured many people would be putting those into the container anyway, so no real need.

If we had more facades, then maybe those as well, but since there's only one, I'm not sure if it is particularly useful.

josephmancuso:
very true. Yeah if people need things loaded in they should just make a service provider to do it.

PR #117, opened 2018-05-08
Add routes command

aisola:
Another command for the new release.

The routes command is simply a convenience craft command which shows you the registered routes in the application.

The routes command will show a table like the following:

```
========  ====================  ===================  ========  ============
Method    Path                  Name                 Domain    Middleware
========  ====================  ===================  ========  ============
GET       /                     welcome                        auth
GET       /login
GET       /logout
POST      /login
GET       /register
POST      /register
========  ====================  ===================  ========  ============
```

This will be especially useful as we add more places to define routes (web.py + api.py, global routes, etc.).

### DIFF
--- a/masonite/commands/RoutesCommand.py
+++ b/masonite/commands/RoutesCommand.py
@@ -1,0 +1,35 @@
+\
+from cleo import Command
+from tabulate import tabulate
+
+
+class RoutesCommand(Command):
+    """
+    List out all routes of the application.
+
+    routes
+    """
+
+    def handle(self):
+        from wsgi import container
+
+        web_routes = container.make('WebRoutes')
+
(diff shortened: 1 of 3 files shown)


aisola:
Okay, I'm not sure why that session test is failing in this PR.

josephmancuso:
Don't worry about the session test. Super annoying. It only fails sometimes. I have to rerun the tests and it passes. Can't figure out why it does that. It never fails on local so it's super hard to debug

masonite/commands/RoutesCommand.py
```diff
@@ -0,0 +1,35 @@
+\
```
josephmancuso: What's this thing? lol

PR #119, opened 2018-05-08
Add optional code reloading into the masonite serve command

aisola:
Yet another command change here for the develop branch. After searching for waitress code reloading, I realized it didn't have any.

I've updated the `craft serve` command to have automatic code reloading. This makes it easier to develop applications in masonite as we won't have to restart the development server in between edits to the code base. You enable code reloading by running `craft serve --reload`. By default, code reloading is turned off.  

The command still uses the waitress server, however, it imports the server rather than calling it out to a subprocess. Using watchdog, it efficiently watches for python files and the `.env` file to change and restarts the server. Updates to HTML and static assets are already reloaded (by design of the Jinja2 and Whitenoise).

There is one downside to automatic code reloading, which is the main reason why I've put this as labelled this as requires discussion. The downside is that the `memory` session driver (default) will break every time the code reloads as the server completely restarts. The obvious work around is to set the session driver to `cookie`.

1. Is this downside something that we can live with?
2. Should we set the session driver to `cookie` by default in 2.0?

### DIFF
--- a/masonite/commands/ServeCommand.py
+++ b/masonite/commands/ServeCommand.py
@@ -16,7 +16,7 @@ class ServeCommand(Command):
     serve
         {--port=8000 : Specify which port to run the server}
         {--host=127.0.0.1 : Specify which ip address to run the server}
-        {--reload : Make the server automatically reload on file changes}
+        {--r|reload : Make the server automatically reload on file changes}
     """
 
     def handle(self):
--- a/masonite/commands/ServeCommand.py
+++ b/masonite/commands/ServeCommand.py
@@ -1,4 +1,11 @@
-from subprocess import call
+
+import os
+import sys
+import time
+import subprocess
+import threading
+
+import waitress
 from cleo import Command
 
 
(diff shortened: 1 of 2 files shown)


josephmancuso:
Let's add waitress as a dependency then we can remove it in 2.0 from the `masoniteframework/masonite` repo.

Super excited to test this out

aisola:
Makes sense. I just realized that waitress wasn't a dependency here. My local tests all passed still. I'll add waitress as a dependency too.

josephmancuso:
as for the session driver, yeah let's go ahead and set it to cookie. the memory driver isn't that great anyway because there are obvious downsides with that. It was basically "use memory in development if you want but deff use cookie in production". The problem with memory is the obvious "what if my site gets 1 million hits overnight." that's like 3-4 million objects being saved in memory which will eventually lead to an obvious memory leak of some order of magnitude. and then when the server crashes everything is lost lol.

so yeah cookie is fine for the driver

aisola:
Looks like this failing test is the session reset.

josephmancuso:
btw the trick I found to developing faster with this repo which you may have noticed is to rename all the directories like `app`, `config`, `resources` etc to `_app`, `_config` and then you can `pip install -e .`

aisola:
Yes, I noticed that earlier today when I was testing out this serve command.

josephmancuso:
you also can't use the memory driver for Gunicorn because memory is not shared between workers so I'm more than happy to set the default driver as cookie

masonite/commands/ServeCommand.py
```diff
@@ -9,14 +16,276 @@ class ServeCommand(Command):
     serve
         {--port=8000 : Specify which port to run the server}
         {--host=127.0.0.1 : Specify which ip address to run the server}
+        {--reload : Make the server automatically reload on file changes}
```
josephmancuso: This is all awesome. The only change I can think of change the `--reload` to `--r|--reload` This will allow us to run `craft serve -r` as well as `craft serve --reload`

josephmancuso: see the ControllerCommand for an example on how cleo handles multiple options

aisola:
There we go, you should be able to use `-r` now as well.

PR #122, opened 2018-05-09
Change request variables

### LINKED ISSUE #121: Request.all() should return a dictionary, not a cgi.FieldStorage
Current Request.all() returns a `cgi.FieldStorage` instance. However, the docs seem to indicate that it returns a dictionary. I think a dictionary is the much more intuitive route for the developers using Masonite.

A field storage instance isn't very difficult to convert to a dict. The following code should suffice.

```python
def field_storage_to_dict(self, field_storage):
        """Get a plain dictionary, rather than the '.value' system used by the cgi module."""
        params = {}
        for key in field_storage.keys():
            params[key] = field_storage[key].value
        return params
```

aisola:
In this PR, is a Proof of Concept for #121. The changes made all pass the tests, but I didn't have time to actually play around with it yet. It should be merge-ready but I should test more first.

`Request.all` should now always return a dictionary. Additionally, the values of each name will be either a `str` (most all), `dict` (from JSON), or `cgi.FormStorage` (if a file). This allows us to have a predictable and intuitive return type from `Request.all` while still maintaining the ability to nicely upload files, etc.

Additional changes in this PR:
* `Request.params` has been renamed to `Request.request_variables` so that new contributors don't confuse it with `Request.url_params` or that class of methods.
* `Request.input` and `Request.has` have been updated to look at new variable.
* `Request.only` added which allows you to request a set variables by name, returning a dictionary of the requested values.

Closes #121


josephmancuso:
Looks good. I'll help test when I get back tonight.

josephmancuso:
Also if you can give just a short code snippet on how it works, I see a few new methods there that aren't `_` or `__` methods

josephmancuso:
if you want, you can recheck out 1.6 and just cherry-pick all these commits you did here and we can push this into 1.6

aisola:
Are we sure we really want this to be put into 1.6? It is technically a breaking change after all. It seems better to put into 2.0 so that the data structures built for 1.6 can stay the same without having 1.6.7 break it.

josephmancuso:
No you're right

josephmancuso:
I'm making the dashboard package and realized the crappy workaround for inputs

aisola:
After my testing, I think this is  ready for merge into develop for 2.0. That's what makes sense to me? Did you still want me to cherry pick the commits out for 1.6? Since it's breaking, it definitely makes sense for 2.0.

josephmancuso:
No 2.0 is good

josephmancuso:
Looks good. It just has some merge conflicts since I've merged 1.6 into develop with another PR

aisola:
Looks like Github wants another review after my update. I can merge after a review.

PR #125, opened 2018-05-10
Added better exception for response type errors

josephmancuso:


josephmancuso:
combining this PR with the #126 PR

PR #133, opened 2018-05-15
Improvements requirements.txt

mapeveri:
I had some problems getting everything. This is my solution.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ bootstrap/cache/*.html
 bootstrap/cache/*.txt
 /uploads/
 .coverage
+/.cache/
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,4 @@
+# Install dependencies masonite
+pip install .
+# Install dev and test dependencies
+pip install -r requirements.txt
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,18 +1,28 @@
 libsass
-Jinja2
-whitenoise
-python-dotenv
-
 validator.py==1.2.5
-cryptography==2.1.4
 bcrypt==3.1.4
 requests==2.18.4
-boto3==1.5.24
 tldextract==2.2.0
-pusher==1.7.4
-ably==1.0.1
+Jinja2==2.10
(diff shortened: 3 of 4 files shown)


josephmancuso:
what does this solve? The only thing in requirements.txt should really be only dependencies for development if there are any.

when testing you should do (preferably inside your virtual environment):

    $ pip install .

which installs everything what would have installed when downloading the actual package. 

or am I missing something?

mapeveri:
I always run pip install -r requirements.txt for develop, In addition there are dependencies that are not in setup.py. We would have to find a way to unify it.

josephmancuso:
If you want to unify it you could create a make command for it

mapeveri:
Make command? I thought to do the following:

Create 3 file requirements. 

1. requirements.txt
2. requirements-test.txt
3. requirements-dev.txt

And in the setup.py read the content of requirements.txt

What do you think?

josephmancuso:
Umm not really a fan. What are some dependencies that are not used in the core package itself. Are there any libraries that are only for development? The solution for this should just be to run `pip install .` which will install all the required packages

josephmancuso:
Just do:

`pip install .`
`pip install -r requirements.txt`

Requirements.txt just needs the ably, pusher, boto3 and coveralls packages

mapeveri:
And now  @josephmancuso what do you think?

josephmancuso:
how does the install.sh work?

mapeveri:
Simplely run this together:

```
pip install .
pip install -r requirements.txt
```

For executed run:

`sh install.sh`

josephmancuso:
hmm I'm sorry. After some thinking I just feel like adding a file to run a test just to combine two commands seems silly to me. This will only be useful when someone adds a package to the repo which really won't be that often.

mapeveri:
It is not only to unify two commands. It is also for:

1. Remove dependencies from requirements.txt and leave only those in setup.py
2. Remove duplicate dependencies in setup.py
3. In requirements.txt leave only the 'dev' and 'tests' dependency.

PR #147, opened 2018-05-26
fixed container resolution resolving similiar objects

josephmancuso:
fixed container resolution.

A bug I found when resolving multiple models loaded into the container is that they were all returning the same model. What the container was doing was searching through the `Country` and `User` model but when annotating them:

```python
from app.User import User
from app.Country import Country

def show(self, user: User, country: Country):
    pass
```

both `user` and `country` were returning the user model. I think this was because `ininstance(user, country)` is true. so it was finding the first instance.

Now what happens is that it will loop through the container and if it finds an instance, it will save it and continue the loop. If nothing else is returned then it will return the instance it found. I'm not entirely sure this is the expected behavior of the container and it can take a longer time to fetch things from the container when resolving instances but it's the best solution I found so far.

Because this is sort of a bug fix, this can go into 1.6.

### DIFF
--- a/masonite/app.py
+++ b/masonite/app.py
@@ -102,8 +102,19 @@ class App():
         """
         Find a given annotation in the container
         """
+        found = False
+        return_class = None
         for provider, provider_class in self.providers.items():
-            if parameter.annotation == provider_class.__class__ or parameter.annotation == provider_class or isinstance(provider_class, parameter.annotation.__class__):
+            
+            if parameter.annotation == provider_class or parameter.annotation == provider_class.__class__:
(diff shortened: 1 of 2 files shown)


josephmancuso:
after the tests are passing, I was able to refactor the method to keep the tests to pass but also clean up a bit and it looks we actually don't need to continue the loop after the object is found. Will be performing more QA With this

josephmancuso:
The solution here was that instead of doing `ininstance`, we can do `issubclass` which is really what we are looking for. We're looking for children classes and not instances of other classes

PR #150, opened 2018-05-27
added init.py loaders for shorthand classes

josephmancuso:
I just added an `__init__.py` file which allows shorthanded imports so this:

```python
from masonite.drivers.UploadDriver import UploadDriver

UploadDriver
<class masonite.driver.UploadDriver.UploadDriver>
```

now can become:

```python
from masonite.drivers import UploadDriver
<class masonite.driver.UploadDriver.UploadDriver>
```

which would have otherwise just imported the module itself before


josephmancuso:
we can also remove the need for the double classes in the providers list so this:

```python
...
'masonite.providers.AppProvider.AppProvider',
'masonite.providers.View.Provider.ViewProvider',
...
```

to

```python
...
'masonite.providers.AppProvider',
'masonite.providers.View.Provider',
...
```

PR #165, opened 2018-06-06
Adds custom error handling

### LINKED ISSUE #51: Add custom 404, 500, 302 etc pages.
There are a few different ways we can do this. One way is to check if a template exists before rendering it. So for example we can check if a `resources/templates/errors/500.html` template exists and render that, else just throw a normal error on the system. This can be done inside the new `ExceptionHandler` class but we will need a new method on the view class:

```python
def show(self):
    view().exists('errors/500')
    # or 
    view().exists('/masonite/snippets/500.html')
```

because then inside our exception class we can do something like:

```python
    def _render(self):
        self._app.bind('StatusCode', '500 Internal Server Error')

        # Check if DEBUG is False
        if not self._app.make('Application').DEBUG or self._app.make('Application').DEBUG == 'False':
            # Raise the exception here
            if self._app.make('View').exists('errors/500'):
                # do all the rendering stuff here
                return
            else:
                raise self._exception

        # return a view
        rendered_view = self._app.make('View')('/masonite/snippets/exception',
            {
                'exception': self._exception,
                'traceback': traceback,
                'tb': sys.exc_info(),
                'app': self._app,
                'enumerate': enumerate,
                'open': open
            }
        ).rendered_template
        self._app.bind('Response', rendered_view)

```

josephmancuso:
Adds support for adding custom error handling. This adds a service provider which renders more professional `404`, `500` etc etc errors.

How it works is that it will look for the error pages in the local application and then default to a masonite error page.

For example if a 500 error is thrown, if the app is in debug mode then it will show the normal debug view. 

If there is a `resources/templates/errors/500.html` template then it will show that.

If there is none then it will show the view inside the masonite package. which simply just shows the error in big text similar to how Laravel shows an error.

Closes #51


josephmancuso:
Needs good tests around the status code provider

PR #167, opened 2018-06-07
added helper to upload provider for static file assets

josephmancuso:
This adds a static file helper to all templates and reads from the drivers in the storage config file. You can do a few things with this PR:

You can specify a `location` in your drivers:

```python
....
's3': {
    's3_client': 'sIS8shn...'
    ...
    'location': 'https://s3.us-east-2.amazonaws.com/bucket'
},
....
```

and then in your templates:

```html
...
<img src="{{ static('s3', 'profile.jpg') }}" alt="profile">
...
```

which simply just returns something like:

```html
...
<img src="https://s3.us-east-2.amazonaws.com/bucket/profile.jpg" alt="profile">
...
```

****

Additionally we may also need multiple locations for a single driver. In that case we can pass in a dictionary:

```python
....
's3': {
    's3_client': 'sIS8shn...'
    ...
    'location': {
        'uploads': 'https://s3.us-east-2.amazonaws.com/uploads',
        'bucket': 'https://s3.us-east-2.amazonaws.com/bucket'
    }
},
....
```

and then access these like:

```html
...
<img src="{{ static('s3.uploads', 'profile.jpg') }}" alt="profile">
or
<img src="{{ static('s3.bucket', 'profile.jpg') }}" alt="profile">
...
```

## Change Log

* Added bug fix to S3 store_prepend driver
* added support for different upload locations with dot notation

You can now do something like:

```python
def show(self, Upload):
    Upload.store(request().input('file'), 'disk.uploads')
```

* Changed the return value of drivers. They now return the file item name and not the location


josephmancuso:
needs test for store prepend on upload driver

Issue #195, opened 2018-06-27
Add default to request input


josephmancuso:
You should be able to do something like:

```python
from masonite.request import Request
...

def show(self, request: Request):
    input = request.input('actual-value', 'default-value')
```

This will return the default value if the actual value does not exist. This will remove the need to do:

```python
from masonite.request import Request
...

def show(self, request: Request):
    if request.has('actual-value'):
        input = request.input('actual-value')
    else:
        input = 'default-value'
```

Should be fairly straight forward. Make sure you ask any questions in [Slack](http://slack.masoniteproject.com)

josephmancuso:
You should just need a test like this to pass:

```
def test_default_request_input_returns_default_value(self):
    self.request.request_variables = {'test1': 'test', 'test2': 'test'}
    assert self.request.input('test3', 'default-value') == 'default-value'
```

(you may need to adjust test slightly). As long as that test passes and all the rest pass then you're golden.

PR #196, opened 2018-06-27
added default value to Request.input()

ChrisByrd14:
#195 

added default value to `Request.input()`

### DIFF
--- a/masonite/request.py
+++ b/masonite/request.py
@@ -41,15 +41,12 @@ class Request(Extendable):
         self.encryption_key = False
         self.container = None
 
-    def input(self, name, default_val=False):
+    def input(self, name, default=False):
         """
         Returns either the FORM_PARAMS during a POST request
         or QUERY_STRING during a GET request
         """
-        if name not in self.request_variables:
-            return default_val
-
-        return self.request_variables.get(name, False)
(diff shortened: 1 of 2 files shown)


masonite/request.py
```diff
@@ -41,11 +41,13 @@ def __init__(self, environ=None):
         self.encryption_key = False
         self.container = None
 
-    def input(self, name):
+    def input(self, name, default_val=False):
```
josephmancuso: This whole thing is perfect but it is such a pet peeve of mine to shorten down variable names. Especially when it's only shortened down by 2 characters lol. Please just make this `default` instead.

masonite/request.py
```diff
@@ -41,11 +41,13 @@ def __init__(self, environ=None):
...
         """
         Returns either the FORM_PARAMS during a POST request
         or QUERY_STRING during a GET request
         """
+        if name not in self.request_variables:
+            return default_val
 
         return self.request_variables.get(name, False)
```
josephmancuso: you know what .... can you actually try just passing the default value into this line right here? That might remove the logic above this line since the False is the default value technically.

josephmancuso:
Great! I'll give it a quick manually check when I get home and merge it into master

PR #202, opened 2018-06-28
Add a way for multiple 'has' checks

nioperas06:
test_compiles_sass's test fails. But I think it has nothing to do with what I added. 😕

### DIFF
--- a/masonite/request.py
+++ b/masonite/request.py
@@ -143,12 +143,14 @@ class Request(Extendable):
     def app(self):
         return self.container
 
-    def has(self, name):
+    def has(self, name, *args):
         """
         Check if a request variable exists
         """
-
-        return name in self.request_variables
+        if args:
+            return name in self.request_variables and all((arg in self.request_variables) for arg in args)
+        else:
+            return name in self.request_variables
 
     def status(self, status):
         self.app().bind('StatusCode', status)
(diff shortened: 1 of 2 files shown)


josephmancuso:
I'm on mobile but looks like it passed good. Dont worry about coverage. Stupid travis CI and PR limitations

masonite/request.py
```diff
@@ -143,12 +143,14 @@ def _get_standardized_value(self, value):
...
-    def has(self, name):
+    def has(self, name, *args):
         """
         Check if a request variable exists
         """
-
-        return name in self.request_variables
+        if args:
+            return name in self.request_variables and all((arg in self.request_variables) for arg in args)
+        else:
```
josephmancuso: So i think there is a bit of redundant coding here. What we should be able to do is have the method signature like:

```python
def has(self, *args):
```

And we don't need the name at all. Whether we pass 1 parameter or 50 it should still work. It would just be a list of 1 element.

tests/test_requests.py
```diff
@@ -46,6 +46,16 @@ def test_request_has_should_return_bool(self):
         assert self.request.has('shouldreturnfalse') == False
 
 
+    def test_request_has_should_accept_multiple_values(self):
```
josephmancuso: This test looks perfect

josephmancuso:
Much cleaner, right?

Issue #197, opened 2018-06-27
Add a way to "pop" off variables from the input.


josephmancuso:
I'm thinking of adding something like:

```python
request.request_variables = {'key1': 'test', 'key2': 'test'}
request.pop('key1') # removes key1

request.request_variables
{'key2': 'test'}
```

This will be useful for the work around of removing the csrf_token from the input

You should be able to pop multiple from a single pop method:

```python
request.request_variables = {'key1': 'test', 'key2': 'test', 'key3': 'test'}
request.pop('key1', 'key2') # removes key1

request.request_variables
{'key3': 'test'}
```

nioperas06:
I'll work on this 👍

PR #206, opened 2018-06-29
Add a way to "pop" off variables from the input. #197

nioperas06:


masonite/request.py
```diff
@@ -387,3 +387,9 @@ def send(self, params):
 
     def helper(self):
         return self
+    
+    def pop(self, *args):
```
josephmancuso: This seems like a lot of extra logic here. Can't the logic just be:

```python
def pop(self, *args):
    for key in args:
        if key in self.request_variables:
            del self.request_variables[key]
```

It looks to me that you are recompiling the whole list and checking if the key in not in the keys. I feel like that will be more expensive and time consuming than the solution above where we are just deleting the keys from the dictionary

tests/test_requests.py
```diff
@@ -325,3 +325,12 @@ def test_request_sets_request_method(self):
         assert request.input('__method') == 'PUT'
         assert request.get_request_method() == 'PUT'
     
+
+    def test_request_has_should_pop_variables_from_input(self):
```
josephmancuso: test looks great

masonite/request.py
nioperas06: Yes, you're rigth.

josephmancuso: you know what ... i don't think this will work since i believe dictionaries are immutable. I wonder if there is another way to just iterate over the popped ones because sometimes there may be 20, 30+ inputs

nioperas06: 😕 This work, I think.

josephmancuso: You could try. I think you will get something along the lines of "dictionary changed during iteration" or something.

josephmancuso: oooh right nvm. It works because we are not iterating over the request variables, we are iterating over the args list. Got it

PR #233, opened 2018-08-08
Add strict mode to container

josephmancuso:
This PR adds a `strict` and `override` settings to the container.

We can now do something like this:

```python
app = App(strict=True)
```

which will throw an exception anytime we try to override a binding (whenever we bind to the same key)

We can also set the override to False

```python
app = App(override=False)
```

This will simply not override the binding in the container. So if we bind twice, the second one will simply just not bind into the container.

This should actually be good for testing because you can provide some default dependencies and then  the container through the loop and the container won't override anything you just set.


aisola:
Looks like the branch is out of date. Merge master in and we should be good to go.

PR #245, opened 2018-08-17
added view based scaffolding

josephmancuso:
This PR adds better view based commands for scaffolding. No more hard to read boiler plate


vaibhavmule:
How about `craft new project-name` should be scaffold this way?

josephmancuso:
umm well craft new is based in the `craft` repo which is sort of design to be a very very minimal CLI tool. With that being said you don't really want to have to always update it. It's more of an install once and you have it sort of thing. If we have a bug fix in the main repo and update craft to have the new bug fix ... and someone doesn't update their craft CLI tool ... Now they keep making new projects with the bug

vaibhavmule:
Woah, You are right. 

That does make sense.

PR #247, opened 2018-08-20
added a way to add route compilers

### LINKED ISSUE #246: Add a way to add simple url compilings
Right now we compile all routes to regex and check for matches. This allows us to check if `/url/@id:int` actually matches `/url/1` and not `/url/joe`.

The "url compilings" (idk what we're gonna call these yet) need to be able to be added to the framework.

In other words we might want to add something like this to our framework:

```
/url/@id:year/@something:slug
```

`year` and `slug` are not actually a thing in the framework and instead of adding them individually, we should be able to add them via a service provider and it should look something like:

```python
def boot(self, Route):
    Route.compile('year', '[0-9]{4}')
```

and then we will be able to use that in our code

josephmancuso:
This PR adds a way to add new route compilers to your application.

I'm currently calling these route compilers but they are the `:int` and `:string` things inside the routes. These compile down to regex in the backend are were previously hard coded. Now they are abstracted away into a property and able to be added through a new `compile()` method.

****

Take this example. We have a route but we only want to accept slugs. Slugs are not available out of the box so we'll need to add a new route compiler:

```python
get('/url/@league:slug')
```

Let's go ahead and boot this into a route class:

```python
def boot(self, Route):
    Route.compile('slug', r'^[a-z0-9]+(?:-[a-z0-9]+)*$')
```

Bam. Now we are able to use `slug` inside our routes. We can also add things like `year` as well or anything else we need:

```python
def boot(self, Route):
    Route.compile('year', r'[0-9]{4}')
```

Note that these need to be regex strings. We COULD in theory accept a normal string and convert it to regex in the backend but I think this is more explicit on explaining what is going on and that we need to accept a regex value.

****

Closes #246


masonite/routes.py
```diff
@@ -7,13 +7,19 @@
...
-from masonite.exceptions import RouteMiddlewareNotFound
+from masonite.exceptions import RouteMiddlewareNotFound, InvalidRouteCompileException
 
 
 class Route:
     """Route class used to handle routing.
     """
 
+    route_compilers = {
+        'int': r'(\d+)',
```
aisola: This is inconsistent. I would change this to `'integer'` based on your other defaults below.

josephmancuso: hmm that would be a breaking change then. We can merge this in and change for 2.1

PR #255, opened 2018-08-29
Remove container resolving

josephmancuso:
This PR that will remove resolving via parameter and only support annotations :cry:
I actually kept the possibility to resolve parameters but it's disabled by default. You can potentially enable it on a per project basis via something like this in your `wsgi.py` file:

```app = App(resolve_parameters=True)```

but it's disabled by default and all Masonite packages and core will only use annotations


vaibhavmule:
What was the reason to remove container resolving?

vaibhavmule:
Thanks for information.

Make sense.

PR #269, opened 2018-09-01
Hotfix/invalid key exception

hammacktony:
Now, the classes throws an InvalidSecretKey exception due to incorrect padding.


masonite/auth/Sign.py
```diff
@@ -38,11 +38,16 @@ def sign(self, value):
...
-        """
 
-        f = Fernet(self.key)
-        self.encryption = f.encrypt(bytes(value, 'utf-8'))
-        return self.encryption.decode('utf-8')
+        Raises:
+            InvalidSecretKey -- Thrown if the secret key does not have correct padding.
+        """
+        try:
+            f = Fernet(self.key)
```
josephmancuso: This is prob more of a pet peeve of mine but can we just wrap the code that was throwing the exception? Can we do something like:

```python
try:
    f = Fernet(self.key)
except binascii.Error:
    raise InvalidSecretKey("The encryption key passed does not have correct padding and therefore cannot be used as a secret key.")

self.encryption = f.encrypt(bytes(value, 'utf-8'))
return self.encryption.decode('utf-8')
```

josephmancuso:
also don't worry about that big drop in test coverage. Some tests don't run on pull requests for security measures

PR #270, opened 2018-09-02
Invalid secret key

hammacktony:
Made sure I branched off of master this time.

This not only raises an `InvalidSecretKey` exception when there is incorrect padding in the secret key. The accompanying test is also included in `tests/test_signing.py`.

`base64.urlsafe_b64decode(key)` is the culprit. This method returns the value of the `base64.b64decode` function, which was where you first saw the error.

It is in the initiation of the Fernet class.

```
class Fernet(object):
    def __init__(self, key, backend=None):
        if backend is None:
            backend = default_backend()

        key = base64.urlsafe_b64decode(key)
        .......
```
It is later found in the decode method in Fernet.

The way it is now catches both occurrences of the error.


tests/test_signing.py
```diff
@@ -28,3 +28,9 @@ def test_sign_without_specifying_key(self):
         s = Sign()
 
         assert s.key == 'NCTpkICMlTXie5te9nJniMj9aVbPM6lsjeq5iDZ0dqY='
+
+    def test_sign_incorrect_padding(self):
+        with pytest.raises(InvalidSecretKey):
+            padded_secret_key = "AQAAQDhAAMAAQAAAAAAAAthAAAAJDczODFmZDM2LTNiOTYtNDVmYS04MjQ2LWRkYzJkMmViYjQ2YQ==="
+            s = Sign(padded_secret_key)
+            assert s.sign('value')
```
josephmancuso: While playing with this i notice that this test didn't pass here that I made to test the new addition:

```python
    def test_sign_incorrect_padding(self):
        with pytest.raises(InvalidSecretKey):
            padded_secret_key = "AQAAQDhAAMAAQYS04MjQ2LWRkYzJkMmViYjQ2YQ==="
            s = Sign(padded_secret_key)
            assert s.sign('value')
```

it was throwing a `ValueError` exception. We can fix this by making something like this in the Sign method:

```python
        except (binascii.Error, ValueError):
            raise InvalidSecretKey("You have passed an invalid secret key of: {}. Make sure you have correctly added your secret key.".format(self.key))
```

this way we catch both exceptions

masonite/auth/Sign.py
```diff
@@ -38,11 +38,18 @@ def sign(self, value):
...
-        f = Fernet(self.key)
-        self.encryption = f.encrypt(bytes(value, 'utf-8'))
-        return self.encryption.decode('utf-8')
+        try:
+            f = Fernet(self.key)
+            self.encryption = f.encrypt(bytes(value, 'utf-8'))
+            return self.encryption.decode('utf-8')
+        except binascii.Error:
+            raise InvalidSecretKey("The encryption key passed does not have correct padding and therefore cannot be used as a secret key.")
+
```
josephmancuso: It's technically OK to just wrap the `f = Fernet(self.key)` key in the try catch and not the decode method since it the `f = Fernet(self.key)` line of code always executes first and therefore will always throw the exception before it hits the decode. If it ever hits the decode method then it can't throw an exception because the secret key is good. Does that make sense?

```python
        try:
            f = Fernet(self.key)
        except binascii.Error:
            raise InvalidSecretKey("The encryption key passed does not have correct padding and therefore cannot be used as a secret key.")

        self.encryption = f.encrypt(bytes(value, 'utf-8'))
        return self.encryption.decode('utf-8')
```

The awesome part about tests is you can restructure your code in anyway you want and as long as the tests pass then you're still good. So doing that small change above still makes the tests pass

masonite/auth/Sign.py
```diff
@@ -3,7 +3,7 @@
 
 from masonite.exceptions import InvalidSecretKey
 
-
+import binascii
```
josephmancuso: Also make sure this is PEP 8 compliant The imports should look like this:

```python
""" Cryptographic Signing Module """

import binascii

from cryptography.fernet import Fernet

from masonite.exceptions import InvalidSecretKey


class Sign:
```

PR #271, opened 2018-09-02
added view tests

josephmancuso:
This PR adds the ability to create "View Tests".

In order to add a view test we just need to add it via a service provider:

```python

def boot(self, view: View):
    view.test('admin', self._is_admin)

def _is_admin(self, user):
    return user.is_admin == 1
```

and then we can use it in our template:

```
{% if user is admin %}
    {{ user.email }} is an admin
{% endif %}
```

Notice we are now able to use the `admin` boolean check in our templates.

A test just passes the obj in question as the first parameter so doing something like:

```html
{% if 'some awesome test' is awesome %}
    {{ user.email }} is an admin
{% endif %}
```

would look something like:

```python

def boot(self, view: View):
    view.test('awesome', self._is_awesome)

def _is_admin(self, string):
    return 'awesome' in string
```

which just looks for the word "awesome" in the object in question. So notice the string was passed in as the first parameter since that is what the test is on.


josephmancuso:
@clsource Need review please ☝️ 😄

tests/test_view.py
```diff
@@ -208,3 +208,25 @@ def test_cache_throws_exception_with_incorrect_cache_type(self):
             view(
                 'test_exception', {'test': 'test'}
             ).cache_for(1, 'monthss')
+
+    def test_can_add_tests_to_view(self):
```
clsource: This test I think does too much. Maybe could be separated in smaller tests

def test_that_admin_user_renders_true()
def test_that_normal_user_renders_false()

josephmancuso: you're right. I sort of snuck it in at the last minute to see if those assertions actually pass. I can refactor as we go though. I'll just need to make a mental note there

PR #285, opened 2018-09-05
Simple bind

hammacktony:
I added the simple method to the App class. This allows you to bind a class into the container by passing the object itself (no need for the key-value pair). I also added a corresponding test in `test_app.py` called `test_app_simple_bind()`. 

Also, I added Pipfile, Pipfile.lock, and requirements.txt for Masonite Core. This allows the user to have an isolated programming development for Masonite core with all the needed packages for testing.

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -23,6 +23,11 @@ class TestApp:
         self.app.simple(Request)
         assert self.app.providers == {'masonite.request.Request': Request}
 
+    def test_app_simple_bind_init(self):
+        req = Request()
+        self.app.simple(req)
+        assert self.app.providers == {'masonite.request.Request': req}
+
     def test_app_makes(self):
         self.app.bind('Request', REQUEST)
         assert self.app.make('Request').cookies == []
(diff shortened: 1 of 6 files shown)


tests/test_app.py
```diff
@@ -18,6 +19,10 @@ def test_app_binds(self):
         self.app.bind('test2', object)
         assert self.app.providers == {'test1': object, 'test2': object}
 
+    def test_app_simple_bind(self):
```
josephmancuso: You should also try to bind an instantiated object as well. I have a sneaky suspicion that it will throw an exception because an instantiated object has no attribute `__name__`

josephmancuso: you might have to do someting like `object.__name__ or object.__class__.__name__` or something

josephmancuso: might look something like:

```python
        self.bind("{module}.{name}".format(
            module=obj.__module__, name=obj.__name__ or obj.__class__.__name__ ), obj)
```

maybe?

hammacktony:
I had a feeling about Request vs Request() and I did not trust my judgement. Good catch! 

That issue should be fixed now. Tests are included as well.

josephmancuso:
I removed the Pipfile since it wasn't really in the right scope of this PR. Let's keep each PR to one thing :)

PR #286, opened 2018-09-06
Crude version of BTree in RouteProvider

hammacktony:
Yesterday we tried a hybrid dictionary/list approach for the RouteProvider to determine if the route is even contained in Masonite's routes library. The idea of [Radix trees](https://en.wikipedia.org/wiki/Radix_tree) came up on Slack. 

[BTrees](https://btrees.readthedocs.io/en/latest/index.html) are a type of tree system akin to Radix trees. I implemented a VERY rough version and it passed the tests. Wanted to let you have a look and toy with it.


theodesp:
I think using Btrees is an overkill. Why don't you try a Radix/ Prefix tree instead?

PR #287, opened 2018-09-06
Update routeprovider

hammacktony:
Minor changes. Cleaned up the code a bit. Also made it read easier. There are two new flagpole comments. Also, created a `create_matchurl` method that cleans up RouteProvider a bit. 

Overall, tried to make RouteProvider easier to read.


Pipfile
```diff
@@ -0,0 +1,27 @@
...
+pendulum = "==1.4.4"
+cleo = "==0.6.5"
+pytest = "*"
+coveralls = "*"
+"Jinja2" = "*"
+
+[dev-packages]
+
+[requires]
+python_version = "3.6"
```
josephmancuso: Let's just remove this Pipfile again. I'd like a single PR with only the Pipfile stuff so I can test only those changes.

masonite/providers/RouteProvider.py
```diff
@@ -127,13 +125,42 @@ def boot(self, route_class: Route, request_class: Request):
...
                     if hasattr(located_middleware, 'after'):
                         located_middleware.after()
+
+                # Breaks the loop because the incoming route is found and executed.
+                # There is no need to continue searching WebRoutes.
                 break
             else:
                 self.app.bind('Response', 'Route not found. Error 404')
+
+    def create_matchurl(self, router, route):
```
josephmancuso: let's put this in the helpers inside `masonite/helpers` directory. you can put it in the `routes.py` file

PR #289, opened 2018-09-06
Added Mail Helper

mitchdennett:


masonite/providers/HelpersProvider.py
```diff
@@ -25,6 +26,7 @@ def boot(self, view: View, request: Request):
         builtins.env = os.getenv
         builtins.resolve = self.app.resolve
         builtins.route = request.route
+        builtins.mail = manager.helper
```
josephmancuso: you know .. after looking at this code I think we're going to have some naming conflicts with the mail module in the project .. Let's name the mail helper `mail_helper()` instead .. 🤔 

sort of ugly but because of how Python manages global namespaces ..

tests/test_managers_mail_manager.py
```diff
@@ -1,4 +1,4 @@
-from config import mail
+from config import mail as mail_config
```
josephmancuso: having to do this sort of draws a red flag about the naming conflicts

mitchdennett: Ahh good point! I will get this changed!

mitchdennett:
@josephmancuso made the change. Check it out

masonite/providers/HelpersProvider.py
```diff
@@ -16,7 +17,7 @@ class HelpersProvider(ServiceProvider):
     def register(self):
         pass
 
-    def boot(self, view: View, request: Request):
+    def boot(self, view: View, request: Request, manager: MailManager):
```
vaibhavmule: I feel it should not be a `manager`, it is quite generic.

josephmancuso: probably right although in the current context it's pretty clear what it's doing. If we need to add a second manager we'll have to refactor a bit

PR #290, opened 2018-09-07
add craft share command

### LINKED ISSUE #205: Create a craft "share" command
This command should simply be a wrapper around the NGROK cli tool and should throw an error if NGROK is not installed

This command should look something like:

```
$ craft share
```

which will fire up an NGROK tunnel to your application

vaibhavmule:
`craft share` command starts a `ngrok` tunnel. 

`craft share` can be run with the following options.
```
  --port=8000 : Specify which port to run the server
  --host=127.0.0.1 : Specify which ip address to run the server
  --path=ngrok : Specify which ngrok executable path to run the ngrok 
```

Example running th command.
`$ craft share --port 5000 --path ~/Downloads/ngrok`

Closes #205

### DIFF
--- a/masonite/commands/ShareCommand.py
+++ b/masonite/commands/ShareCommand.py
@@ -1,0 +1,30 @@
+from cleo import Command
+import shutil
+import subprocess
+
+
+class ShareCommand(Command):
+    """
+    Strats a ngrok tunnel
+
+    share
+        {--port=8000 : Specify which port to run the server}
+        {--host=127.0.0.1 : Specify which ip address to run the server}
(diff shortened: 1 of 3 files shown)


vaibhavmule:
@josephmancuso Should I base off this from a master as it is next-minor and currently it is based off from develop

josephmancuso:
develop is fine. We'll just move this to the next major

vaibhavmule:
Exactly,

Creating wrapper over other command is kind of task, also you have to make sure get all the user input.

I feel we should not add wrapper over `ngrok` command, simply opening a new terminal and running a command is better. 

We should just get rid of issue #205 and close this pull request.

PR #296, opened 2018-09-08
Remove the redundant coveralls reporting

### LINKED ISSUE #291: Remove the redundant coveralls reporting on every single python version
You may have noticed that coveralls comments like 3 or 4 times on a pull request reporting the coverage. This is because it is set to report on every successful build in the travis.yml file. We need to pick 1 python version and only report for that

vaibhavmule:
Closes #291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,6 @@ deploy:
 
 jobs:
   include:
-    - stage: "coveralls"
-      python: "3.6"
+    - stage: coverage
+      python: 3.6
       script: coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,5 +27,8 @@ deploy:
     tags: true
     python: 3.6
 
-after_success:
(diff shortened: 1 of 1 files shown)


vaibhavmule:
@coveralls Why is it showing coverage:unknown and not showing percentage?

josephmancuso:
Hm looks like the coverage ran but it didn’t return a percentage?

vaibhavmule:
@josephmancuso it may be because this commit does not require any coverage and we had omitted `.travis.yml` in https://github.com/MasoniteFramework/core/blob/master/.coveragerc

josephmancuso:
i'll merge this in and hope it works 🤞

vaibhavmule:
Got a reply, let me write it again and check if it works. 
https://github.com/lemurheavy/coveralls-public/issues/1185#issuecomment-420083282

PR #306, opened 2018-09-08
Added Craft Test Command

hammacktony:
Added new craft command. Allows the user to easily test their application. 

Instead of using `python -m pytest`, one can use `craft test`.


josephmancuso:
hmm although this is sort of useful .. 🤔 I'm not sure if it's actually useful. it's really only saving you from typing the `-m` part. You're typing 2 words instead of 3 words.

theodesp:
A more useful command would be `craft make:test CustomerTest`. This will create a basic test example

PR #308, opened 2018-09-09
added several methods to the service provider class

josephmancuso:
This PR adds several new methods onto Service Providers.

We no longer need to remember the weird syntax related to how Masonite finds these in the container. now it's as simple as:

```python
def boot(self):

    self.routes([
        get() ..
    ])

    self.http_middleware([Middleware, Here])

    self.route_middleware({'key': Middleware})

    self.migrations('directory1', 'directory/2')

    self.commands(Command1(), Command2())

    self.assets({'/alias': 'directory/})
```


masonite/provider.py
```diff
@@ -38,3 +41,40 @@ def load_app(self, app):
...
+        http_middleware += middleware
+
+    def route_middleware(self, middleware):
+        route_middleware = self.app.make('RouteMiddleware')
+        route_middleware.update(middleware)
+
+    def migrations(self, *directories):
+
+        for directory in directories:
+            random_string = ''.join(random.choice(
```
clsource: I think this could be refactored in a helper method like this

```python

import hashlib
import uuid
import string
import random


def random_string(length=4):
    rnd_string = "".join(
        random.choice(string.ascii_uppercase + string.digits) for _ in range(length)
    )
    return rnd_string


def random_hash():
    uid = uuid.uuid4().bytes
    return hashlib.sha256(uid).hexdigest()


print(random_string())

print(random_hash())

```

masonite/provider.py
```diff
@@ -38,3 +39,62 @@ def load_app(self, app):
...
+                directory
+            )
+
+    def commands(self, *commands):
+        """Adds commands to the container. Pass in the commands as arguments.
+        """
+
+        for command in commands:
+            self.app.bind(
+                '{}Command'.format(random_string(4)),
```
clsource: the default value is 4 so `random_string()` should work 👍.
nevertheless a explicit value adds readability

vaibhavmule:
This is awesome and cleaner for web devs. 

Good abstraction layer.

PR #310, opened 2018-09-09
Craft test with command flag

hammacktony:
Mentioned on Slack about the ability to add a flag to `craft serve` where it would run the tests then serve application. It also works with the automatic reloading.

`craft serve -t` to run test then serve application.
`craft serve -r -t` to run tests each time the server auto reloads the application.


hammacktony:
I don't know. 

Some tests are not that critical. I mean, while testing Masonite-core, the cache driver sorta passes. But is that critical?

It would be easy to stop the application, but it is your call.

josephmancuso:
fair enough. This is actually really freakin cool. Not only are we running tests with the server on reload but every single time a file is saved / changed we can ensure the tests still pass so we can keep modifying our application until tests pass. 

Awesome job I really love this 🏅

masonite/commands/ServeCommand.py
```diff
@@ -68,6 +74,26 @@ def _run_with_reloader(self, extra_files=None, interval=1):
...
+    def _run_with_reloader_and_test(self, extra_files=None, interval=1):
+        """Runs pytest before running the given function in an independent python interpreter."""
+        import signal
+        from wsgi import application
+        reloader = WatchdogReloaderLoop(
+            extra_files, interval, log_func=self.comment)
+        signal.signal(signal.SIGTERM, lambda *args: sys.exit(0))
+        try:
+            if os.environ.get('MASONITE_RUN_MAIN') == 'true':
+                subprocess.call(['python', '-m', 'pytest'])
```
josephmancuso: Maybe we should do something here like:

```python
if self.option('test') == 'pytest':
    subprocess.call(['python', '-m', 'pytest'])
else:
    subprocess.call(self.option('test').split(' '))
    # or 
    subprocess.call(self.option('test'), shell=True)
```

hammacktony:
After reading blogs and listening to podcasts, it seems that Pytest is the gold standard these days. I know about Nose and Tox, but never used them.

Would there be a reason to have a different tester?

clsource:
I think pytest is fine since Masonite use it by default.
If someone wants to use green https://github.com/CleanCut/green
or similar maybe they could provide a testing command in an env variable

`$ export MASONITE_TEST_RUNNER=green` 

IDK is just an idea 👍

hammacktony:
Joseph and I sort of changed the goal of this project. We realized that it would be beneficial to not pigeon hole our selves with the testing suites. Instead, we are going to broaden the concept. 

Have `--c|command` flag enable the user to run a command before the server starts. It also works with autoreload enabled.

Ex. `craft serve -c "python -m pytest"` and `craft serve -r -c "python -m pytest"`

This allows one to run any command and implement and test suite. 

Thoughts?

clsource:
Seems cool.
Although having a default pytest command built in would be helpful too.

`craft serve -r -c "python -m pytest"` could become craft serve -r --pytest

hammacktony:
Yeah. Pytest is easy. 

But what if someone wants to use nose2 or tox? I don’t know why anyone would use any other testing frameworks. But, this allows for more flexibility when a user is working on their project. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

clsource:
Yes I agree that the `-c|command` option is needed. I just say that since pytest is so common and used in masonite by default. Including a built in command for that use case would be cool too 👍

clsource:
what about this. 
c|command option will execute inline command
`craft serve -r -c "python -m pytest"`

s|script option will execute a script. 

`craft serve -r -s pytest`

this will look for a file named `pytest` inside the current working directory.

hammacktony:
I don’t want to sound like a jerk, but why would we want that? Can you give an example of how it could be beneficial? 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

clsource:
Nah it's good. I realise is the same thing xd.

clsource:
Nah it's good. I realise is the same thing xd.

josephmancuso:
is this ready for a review?

hammacktony:
Yeah. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

josephmancuso:
🤔 hmm i pulled in the changed but when i run the command:

screen shot 2018-09-11 at 10 45 48 pm

it says too many arguments? How do you use the command?

hammacktony:
Just like that. I forgot to commit something. It should be this:
`{--c|command=None : Test project before running the server}`

Pushing commit in a second.

PR #317, opened 2018-09-10
Maintenance mode middleware

### LINKED ISSUE #282: Implement Down for Maintenance Mode
It would be useful to provide this functionality:

https://laravel.com/docs/5.7/configuration#maintenance-mode

vaibhavmule:
Closes #282


masonite/middleware/MaintenanceModeMiddleware.py
```diff
@@ -0,0 +1,21 @@
...
+        ''' Inject Any Dependencies From The Service Container '''
+        self.request = Request
+
+
+    def before(self):
+        ''' Run This Middleware Before The Route Executes '''
+        down = os.path.exists(os.path.join(application.BASE_DIRECTORY, 'bootstrap/down'))
+        if down is True:
+            self.request.status('503 Service Unavailable')
+            return # todo return response request lifecycle should stop
```
vaibhavmule: @josephmancuso 

I got stuck what to return on this line.

josephmancuso: may not have to return anything if that works. Middleware is really just designed to changed some configurations before reaching (or not reaching) the view

josephmancuso: but do use double quotes now instead of single quotes. 2.1 will be more closer to "PEP 8" standards than 2.0 was

vaibhavmule: Not working as expected, I was testing with the freshly installed project via `craft new hello` and it was always showing that welcome page even after this middleware is implemented.

vaibhavmule:
Closing this PR as currently, Request Lifecycle goes through the various service provider and it does require https://github.com/MasoniteFramework/masonite/blob/master/bootstrap/start.py#L43 `loop` to be a break after discussing with @josephmancuso said, it is not a good idea to break the loop.

josephmancuso:
no we don't need to break the loop

josephmancuso:
we just need to set the status code in the middleware

josephmancuso:
if we wanted to break when the middleware was set and simply return the status code of whatever the middleware was we would need to do the conversation we had about breaking from the provider loop but setting a status code in a provider should be good enough

vaibhavmule:
Ok, My bad, misunderstanding,

but the status code gets changed when it goes to route after the middleware. 

You mean specify status code in `after()`?

vaibhavmule:
it didn't require to change `RouteProvider`, moving code to `after()` works and will show this.
screen shot 2018-09-13 at 10 22 46 am

josephmancuso:
what is the status of this? is this ready for a review?

vaibhavmule:
Sorry, I need to write tests for it and will update you.

josephmancuso:
👍

vaibhavmule:
@josephmancuso it is ready for review.

PR #318, opened 2018-09-10
HTTP status mapping int for str

### LINKED ISSUE #314: Create a status map and allow integers to be returned as the status
right now the request status is very explicit:

```python
request.status('200 OK')
```

which is great because it is very explicit but if we want to return a permanent redirection we would need to do:

```python
request.status('301 Moved Permanently')
```

which means we can't typo or it will return a weird response. So we should be able to do something like:

```python
request.status(301)
```

What this will do is do some lookup in a dictionary key map that we put somewhere and return the correct response. Something like this:

```python
statuses = {
    301: '301 Moved Permanently' ,
    ...
}

def status(self, status):
    if ininstance(status, str):
        # no new logic, keep the logic as is.
    elsif instance(status, int):
        # lookup via the status map
        self.app().bind('StatusCode', statuses[status])
```

this is only the general concept. We can put the keymap in a `masonite/helper` file somewhere and then we can import it into the request class and maybe put it on the request class itself:

```python
from masonite.helpers.something import response_statuses
..
class Request:

    status = response_statuses()
```

maybe something like that?

vaibhavmule:
Closes #314

add status code based on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status) excepts deprecated one and add [103](https://tools.ietf.org/html/rfc8297)

### DIFF
--- a/masonite/helpers/status.py
+++ b/masonite/helpers/status.py
@@ -1,0 +1,81 @@
+""" Helper Functions for working with Status Codes"""
+
+statuses = {
+    100: '100 Continue',
+    101: '101 Switching Protocol',
+    102: '102 Processing',
+    103: 'Early Hints',
+    200: '200 OK',
+    201: '201 Created',
+    202: '202 Accepted',
+    203: '203 Non-Authoritative Information',
+    204: '204 No Content',
+    205: '205 Reset Content',
+    206: '206 Partial Content',
+    207: '207 Multi-Status',
+    208: '208 Multi-Status',
+    226: '226 IM Used',
+    300: '300 Multiple Choice',
+    301: '301 Moved Permanently',
+    302: '302 Found',
+    303: '303 See Other',
+    304: '304 Not Modified',
+    307: '307 Temporary Redirect',
+    308: '308 Permanent Redirect',
+    400: '400 Bad Request',
+    401: '401 Unauthorized',
+    402: '402 Payment Required',
+    403: '403 Forbidden',
+    404: '404 Not Found',
+    405: '405 Method Not Allowed',
+    406: '406 Not Acceptable',
+    407: '407 Proxy Authentication Required',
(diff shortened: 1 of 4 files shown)


masonite/helpers/status.py
```diff
@@ -0,0 +1,87 @@
...
+    505: '505 HTTP Version Not Supported',
+    506: '506 Variant Also Negotiates',
+    507: '507 Insufficient Storage',
+    508: '508 Loop Detected',
+    510: '510 Not Extended',
+    511: '511 Network Authentication Required',
+}
+
+
+def response_statuses(self, status):
```
josephmancuso: this is very strange 🤔 
It's like a detached bound method .. 

this is a helper and shouldn't have `self` as a parameters. Bound methods are methods and should go inside classes.

Make the logic of this inside the `status()` method of the request class and make this helper function simply return the dictionary above.

masonite/request.py
```diff
@@ -29,6 +30,7 @@ class Request(Extendable):
         Extendable {masonite.helpers.Extendable.Extendable} -- Makes this class have the ability to extend another class at runtime.
     """
 
+    status = response_statuses
```
josephmancuso: do `response_statuses()` (as a function) which should return the dictionary of status code mappings

masonite/request.py
```diff
@@ -241,19 +243,6 @@ def has(self, *args):
 
         return all((arg in self.request_variables) for arg in args)
 
-    def status(self, status):
```
josephmancuso: This should be added back

vaibhavmule: @josephmancuso I have made the commit, 

Even while implementing I thought it is weird, I thought helper in another way.

vaibhavmule: Where should I put `KeyError` exception?

PR #322, opened 2018-09-10
Cors middleware

hammacktony:
Adds ability to inject CORS headers into HTTP Response.

The corresponding test is included.


masonite/middleware/CorsMiddleware.py
```diff
@@ -0,0 +1,24 @@
...
+        |--------------------------------------------------------------------------
+        |
+        '''
+
+        # Demo CORS values - Modify for your own CORS values
+        cors = {"Access-Control-Allow-Origin": "*",
+                "Access-Control-Allow-Headers": "Content-Type",
+                "Access-Control-Allow-Methods": "DELETE, GET, HEAD, OPTIONS, PATCH, POST, PUT"}
+
+        for key, value in cors.items():
```
josephmancuso: Theres just something about this I don't like 🤔 

it looks very dirty

josephmancuso: i'm gonna update the headers method real quick to accept a dictionary

josephmancuso: https://github.com/MasoniteFramework/core/pull/339

josephmancuso: ok now we should be able to do something like:

```
self.request.header({"Access-Control-Allow-Origin": "*",
                "Access-Control-Allow-Headers": "Content-Type",
                "Access-Control-Allow-Methods": "DELETE, GET, HEAD, OPTIONS, PATCH, POST, PUT"})
```

a little cleaner

hammacktony:
I am with you on all of those changes. However, I have a slight syntax issue.

```
from masonite.middleware import CorsMiddleware as Middleware


class CorsMiddleware(Middleware):
```
I think the naming is confusing. You have two `CorsMiddleware`. It is not....intuitive I guess. Besides this, I really like your changes.

josephmancuso:
well technically it is 1 middleware. Since you are importing as a different namespace

josephmancuso:
why was this closed again?

PR #324, opened 2018-09-11
Added route views

### LINKED ISSUE #294: View Routes
It would be useful to provide this functionality 🤔 :

https://laravel.com/docs/5.6/routing#view-routes

josephmancuso:
This PR adds route views. We can use them like this:

```python
ROUTES = [
    Get().view('/url/to/something', 'template/location', {'key': 'value'})
]
```

This will simply return the template instead of needing a controller at all

Closes #294


josephmancuso:
I also removed all the flag pole comments because it was messing up the reference docs when generating those

josephmancuso:
affect the code how so? in terms of being similiar?

clsource:
Not affect the code since they are just comments. But the docs. Are there other places that flagpole could damage rendering the docs?. Idk I just though about that.

josephmancuso:
oh ummm . I used to be a fan of flagpole comments but my thoughts have changed overtime lol. I think the only place I put them were in service providers and in the `/masonite` repo itself

clsource:
It has nothing to do with this pull request though. Just maybe look for another files that may need to ditch that comment style

clsource:
It has nothing to do with this pull request though. Just maybe look for another files that may need to ditch that comment style

PR #326, opened 2018-09-11
Remove redundant coveralls report

### LINKED ISSUE #291: Remove the redundant coveralls reporting on every single python version

vaibhavmule:
Closes #291 

Reference to comment: https://github.com/lemurheavy/coveralls-public/issues/1185#issuecomment-420083282


vaibhavmule:
Removed it.

PR #328, opened 2018-09-12
added hotfix for multi select values

josephmancuso:
This PR fixes an issue where multi select dropdowns were not able to be fetched via the input correctly.

Previously something like this:

```python
<form action="/test" method="POST">
    {{ csrf_field|safe }}
    <select name="size_list[]" multiple class="form-control">
        <option value="1">This is a value</option>
        <option value="2">This is a value</option>
        <option value="3">This is a value</option>
    </select>

    <select name="size_list[testing]" multiple class="form-control">
        <option value="1">This is a value</option>
        <option value="3">This is a value</option>
    </select>
    <input type="submit">
</form>
```

Would not actually send all the selected values to the input backend. It would only send something like:

```python
{'size_list[]' MiniFieldStorage(key, 1) ... }
```

something like that. Now it will do something like:

```python
{'size_list': 1}
```

or

```python
{'size_list': [1, 2]}
```

depending on the amount of elements in the select.

Also notice that an input name of `size_list[]` is actually fetched via `size_list` and not the `size_list[]`. This is open for debate.

```python
request.input('size_list') # [1,2,3]
```

also this:

```python
    <select name="size_list[testing]" multiple class="form-control">
        <option value="1">This is a value</option>
        <option value="3">This is a value</option>
    </select>
```

will be fetched via `size_list[testing]`:

```python
request.input('size_list[testing]') # [1,2,3]
```


sidneywidmer:
Wouldn't something like `request.input('size_list.testing')` to access nested arrays be a little nicer? Plus we're already using dot notation in other parts of the framework so it would be a little more consistent

tzsk:
I have created a `Repl` Project in `Express` to better understand how the Request Parsing works.

https://repl.it/@tzsk/RequestParsing

**NOTE:** Please go inside `views/index.twig` to look at the template [multi-select fields].

josephmancuso:
@sidneywidmer latest commit adds that functionality. We can do:

```python
request.input('something.else')
```

which is exactly the same as doing:

```python
request.input('something[else]')
```

PR #334, opened 2018-09-14
added base support for using full routes

### LINKED ISSUE #329: URL Generation
It would be useful to provide this functionality : https://laravel.com/docs/5.6/urls
Example:
```
group([
        get('/activate/@token', 'RegisterController@validate').name('validate')
    ], prefix='/auth', name='auth.'),
```
Before
```
Notify.mail(EmailVerificationNotification, to=user.email, link='http://localhost:8000/activate/{0}'.format(token))
```
After
```
Notify.mail(EmailVerificationNotification, to=user.email, link=route('auth.activate', ['token' => token]))
```

josephmancuso:
This adds support for adding "full" urls using the route method on the request class.

A code example would be this:

```python
def show(self, request: Request):
    request.route('dashboard.user') # /dashboard/user
    request.route('dashboard.user', full=True) # http://localhost:8000/dashboard/user
    request.route('dashboard.user', {'id': 1}) # /dashboard/user/1
    request.route('dashboard.user', {'id': 1}, full=True) # http://localhost:8000/dashboard/user/1
    request.route('dashboard.user', {'id': 1}, True) # http://localhost:8000/dashboard/user/1

```

Closes #329

### DIFF
--- a/masonite/request.py
+++ b/masonite/request.py
@@ -15,11 +15,12 @@ from urllib.parse import parse_qs
 import tldextract
 from cryptography.fernet import InvalidToken
 
+from config import application
 from masonite.auth.Sign import Sign
+from masonite.helpers import dot
 from masonite.helpers.Extendable import Extendable
 from masonite.helpers.routes import compile_route_to_regex
 from masonite.helpers.time import cookie_expire_time
-from masonite.helpers import dot
 
 
 class Request(Extendable):
@@ -533,8 +534,6 @@ class Request(Extendable):
             self
         """
 
-        web_routes = self.container.make('WebRoutes')
-
         self.redirect_url = self._get_named_route(route_name, params)
 
         return self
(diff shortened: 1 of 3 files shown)


josephmancuso:
@nioperas06 ☝️

masonite/request.py
```diff
@@ -613,7 +612,8 @@ def route(self, name, params={}):
             masonite.routes.Route|None -- Returns None if the route cannot be found.
```
clsource: Needs to specify the full=False param and what is it purpose 👍

clsource: Also adding a convenience function with the full=True would be helpful

```python
def route_full(self, name, params={}):
  return self.route(self, name, params, full=True)
```

josephmancuso: good catch on the doc string. I'm not too crazy about wrapping the method in another method though

PR #346, opened 2018-09-19
Route with and without trailing slash resolve the same content

### LINKED ISSUE #292: Why is not router just show same content for '/blog/' and '/blog'?
I have followed tutorial about creating a blog in `masonite`,  I wrote a route which points to `/blog` but when I visit `/blog/` it does not return the same. 
if it is not implemented, I would suggest this to have in masonite as [usually trailing slash and non-trailing slash URLs often serve the same content and it just that you have to have preferred version](https://webmasters.googleblog.com/2010/04/to-slash-or-not-to-slash.html)

vaibhavmule:
Closes #292


masonite/helpers/routes.py
```diff
@@ -178,8 +178,14 @@ def create_matchurl(router, route):
...
+        else:
+            matchurl = re.compile(regex.replace(r'\/\/$', r'$'))
     else:
-        matchurl = re.compile(regex.replace(r'\/$', r'$'))
+        if router.url.endswith('/'):
+            matchurl = re.compile(regex)
+        else:
+            matchurl = re.compile(regex.replace(r'\/$', r'$'))
 
     return matchurl
```
josephmancuso: 🤔 I'll have to play with this when I pull it in. You might be able to condense this logic a bit.

Sort of a note to myself when I test this

PR #350, opened 2018-09-19
refactor code as per pep8 using flake8

vaibhavmule:
used `flake8 masonite/ --ignore=E501,F401,E128,E402,E731,F821,E712 `


masonite/providers/AppProvider.py
```diff
@@ -73,9 +73,9 @@ def boot(self, request: Request, route: Route):
...
         Autoload(self.app).load(directories)
-    
+
     def _set_application_debug_level(self):
-        if self.app.make('Application').DEBUG == 'True':
-            self.app.make('Application').DEBUG == True
-        elif self.app.make('Application').DEBUG == 'False':
-            self.app.make('Application').DEBUG == False
+        if self.app.make('Application').DEBUG is 'True':
+            self.app.make('Application').DEBUG is True
```
vaibhavmule: @josephmancuso Your opinion?

josephmancuso: or is that not true?

josephmancuso: nvm. That's not true. yeah i'm a fan of doing it like that

vaibhavmule: This is what best explanation I can find. 
`is` checks that 2 arguments refer to the same object, `==` checks that 2 arguments have the same value.

So, I need to change that back to `==` for this line `if self.app.make('Application').DEBUG is 'True':`

josephmancuso: ok that's what I thought lol

PR #352, opened 2018-09-19
add craft middleware command

### LINKED ISSUE #293: craft middleware MiddlewareName : Create a new middleware

vaibhavmule:
Closes #293


masonite/commands/MiddlewareCommand.py
```diff
@@ -0,0 +1,36 @@
...
+import os
+
+from masonite.view import View
+from masonite.app import App
+from masonite.helpers.filesystem import make_directory
+
+from cleo import Command
+
+
+class MiddlewareCommand(Command):
```
josephmancuso: This should look more like this file https://github.com/MasoniteFramework/core/blob/develop/masonite/commands/ModelCommand.py

masonite/snippets/scaffold/middleware.html
```diff
@@ -0,0 +1,17 @@
+""" {{ class }} Middleware """
+
+class {{ class }}Middleware:
+    """{{ class }} Middleware
+    """
+
+    def __init__(self, Request):
```
josephmancuso: 2.1 doesn't support Parameter Resolving anymore. You'll need to do:

```python
...
def __init__(self, request: Request):
```

masonite/snippets/scaffold/middleware.html
```diff
@@ -0,0 +1,17 @@
...
+        """ Inject Any Dependencies From The Service Container """
```
josephmancuso: Again follow the docstring here like in the core repo. Example here https://github.com/MasoniteFramework/masonite/blob/develop/app/http/middleware/AuthenticationMiddleware.py#L10-L15

masonite/snippets/scaffold/middleware.html
```diff
@@ -0,0 +1,17 @@
...
+        self.request = Request
+
+    def before(self):
+        """ Run This Middleware Before The Route Executes """
+        pass
+
+    def after(self):
+        """" Run This Middleware After The Route Executes """
```
josephmancuso: docstring is messed up. Has too many `""""` Also let's follow the docstrings I've been using in other files:

```python
    def after(self):
        """Run This Middleware After The Route Executes
        """
```

masonite/commands/MiddlewareCommand.py
josephmancuso: 🤔 no lol. Maybe we should add that way in this pr?

Just need to take a look at the BaseScaffoldCommand and maybe add something like:

```
...
postfix='Middleware'
```

and then we can add that to scaffolding commands to postfix stuff.

josephmancuso: you'll need to add the logic to append file names in that `BaseScaffoldCommand`.

vaibhavmule: How about `suffix` true.

```
...
suffix = True
```

masonite/snippets/scaffold/middleware.html
vaibhavmule: @josephmancuso Made the changes as requested.

vaibhavmule:
Oh pep8, let me fix it that.

PR #356, opened 2018-09-21
Add `port` dictionary key

bjorntheart:
This kept me busy for the past hour trying to connect to my Postgres instance. I kept getting connection errors when trying to `craft migrate`. I assumed the environment varialbe `DB_PORT` was included in the `config/database.py`, but it wasn't
I added support for non-standard database ports to `config/database.py`

### DIFF
--- a/config/database.py
+++ b/config/database.py
@@ -32,6 +32,7 @@ DATABASES = {
     'default': {
         'driver': os.environ.get('DB_DRIVER'),
         'host': os.environ.get('DB_HOST'),
+        'port': os.environ.get('DB_PORT'),
         'database': os.environ.get('DB_DATABASE'),
         'user': os.environ.get('DB_USERNAME'),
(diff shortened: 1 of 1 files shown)


josephmancuso:
What the heck? How is this forgotten? lol. Can you add this to this file as well in the `/masonite` repo https://github.com/MasoniteFramework/masonite/blob/master/config/database.py. branch off master (can prob do the whole thing through GitHub interface)

josephmancuso:
🤔 i've definitely connected to postgres though. Is it a non standard port? Maybe orator defaults to 5432 if no port is set.

bjorntheart:
Yeah, orator defaults to 5432. I'm using 54322, because I have multiple Postgres containers running on my development machine

josephmancuso:
ahh good catch

josephmancuso:
that was probably a really annoying issue

PR #360, opened 2018-09-22
Add pipfile

josephmancuso:


Pipfile
```diff
@@ -27,3 +27,4 @@ orator = "==0.9.7"
 
 [dev-packages]
 waitress = "==1.1.0"
+masonite-cli = "*"
```
vaibhavmule: also add, 
```
pytest = "*"
coveralls = "*"
```

josephmancuso:
let's close these down. If we are going to go with pipenv let's try to start fresh again

PR #361, opened 2018-09-23
do the spring cleaning

vaibhavmule:
Not completed yet.


config/middleware.py
```diff
@@ -1,18 +1,52 @@
...
+| HTTP middleware is middleware that will be ran on every request. Middleware
+| is only ran when a HTTP call is successful (a 200 response). This list
+| should contain a simple aggregate of middleware classes.
+|
+"""
 
+HTTP_MIDDLEWARE = [
+    LoadUserMiddleware,
+    # todo
+    # CsrfMiddleware,
```
vaibhavmule: Tests were failing, So commented out `CsrfMiddleware`

vaibhavmule:
Not sure about, what to do with `.env` files.?

josephmancuso:
ok when i get the chance I'll pull this in and try to fix anything. They could be real test failures

vaibhavmule:
sure, also, the `Storage().compile_sass()` wasn't working as expected, I'm not sure if I miss something for sass compilation.

josephmancuso:
just needs the libsass package

vaibhavmule:
After doing `pip install libpass`, compile thing is fixed. Look like I haven't done `pip install -r requirements.txt` in a while.

josephmancuso:
so annoying how you can't commit to pull requests but travis throws a crap ton of errors. in .gitignore there is this line: `storage/compiled/`

try uncommenting that out and committing that `style.css` file

vaibhavmule:
Good catch, I have committed `style.css`.

vaibhavmule:
Tests are failing here, https://github.com/MasoniteFramework/core/blob/develop/tests/testing/test_unit_test.py#L59

config/middleware.py
vaibhavmule: All tests are running, but `CsrfMiddleware` still commented.

.gitignore
```diff
@@ -11,7 +11,7 @@ htmlcov
 .env
 .pytest_cache/
 __pycache__
-storage/compiled/
+# storage/compiled/
```
vaibhavmule: for now I have commented it out, I can remove this if needed.

PR #362, opened 2018-09-23
fix Job Command

vaibhavmule:


vaibhavmule:
We should have tests for all of the commands in `core` and `craft` repos.

josephmancuso:
yeah i believe there was someone working on being able to test these

josephmancuso:
the problem is a lot of them scaffold the application out and once we have #361 merged we might be able to make this easier

PR #365, opened 2018-09-24
Email Verification Feature

mitchdennett:
Still need to add tests and documentation.


masonite/auth/MustVerifyEmail.py
```diff
@@ -0,0 +1,24 @@
...
+from masonite.managers import MailManager
+from masonite.request import Request
+
+
+class MustVerifyEmail:
+    """Class To Verify User Email
+    """
+
+    def verify_email(self, mail_manager: MailManager, request: Request):
+        mail = mail_manager.helper()
```
josephmancuso: why do `.helper()`. That just returns `self`

masonite/snippets/auth/controllers/RegisterController.py
```diff
@@ -5,7 +5,7 @@
 from masonite.helpers import password as bcrypt_password
 from masonite.request import Request
 from masonite.view import View
-
+from masonite.auth.MustVerifyEmail import MustVerifyEmail
```
josephmancuso: add this authentication class to the __init__.py file so we can just do `from masonite.auth import MustVerifyEmail` instead of this longer more redundant form.

masonite/middleware/VerifyEmailMiddleware.py
```diff
@@ -0,0 +1,17 @@
...
+        self.request = request
+
+    def before(self):
+        user = self.request.user()
+        if(user.verified_at is None){
+            self.request.redirect('/email/verify')
+        }
+
+    def after(self):
+        pass
```
josephmancuso: Just need to add the better docstrings to this too

mitchdennett: I actually just need to remove this file. It should just be in the Masonite repo right.

PR #366, opened 2018-09-25
added match routes

josephmancuso:
This PR adds the ability to have multiple request methods on a single route by introducing new "Match routes".

These look like this:

```python
from masonite.routes import Match

ROUTES = [
    Match(['GET', 'POST']).route('...', '...')
]
```

which will match both GET and POST requests


clsource:
Maybe adding constants would be helpful 

from masonite.routes.constants import GET, POST

vaibhavmule:
How about just using our current GET(), and POST() methods?

```
from masonite.routes import Match, Get, Post

ROUTES = [
    Match([Get(), Post()]).route('...', '...')
]
```

josephmancuso:
@clsource for spaces, yes it likely would since `POST` in `POST ` (<-- space) is True

As for lowercase .. 🤔 .. interesting. Since maybe not everyone knows that it needs to be in all uppercase. I'll make that change!

@vaibhavmule 🤔 
```
from masonite.routes import Match, Get, Post

ROUTES = [
    Match([Get(), Post()]).route('...', '...')
]
```

i think that needlessly complicates things. Just doing basic strings should be the way to go. I don't see the benefit in adding the extra hoops to do that but interesting thought!

clsource:
Maybe something like this 
if method.lower().strip() in methods:
   valid

josephmancuso:
@clsource well we have to make this work:

```python
request.get_request_method() in route.method_type
```

so a better approach would probably be to just adjust the incoming list in the constructor of the Match class:

```python
    def __init__(self, method_type=['GET']):
        """Post constructor
        """
         self.method_type = [x.upper() for x in method_type]
        self.list_middleware = []
```

or something like that. So just forcing all the request methods in the list to be uppercase since the incoming request method is always uppercase

josephmancuso:
@MasoniteFramework/reviewers Made appropriate changes. Now you MUST pass in a list:

```
Match(['GET', 'POST']).route('..')
```

Anything other data type will throw an exception.

Also made the ability to use different type cases:

```
Match(['Get', 'Post']).route('..')
```

This will be converted to uppercase

clsource:
@josephmancuso maybe using https://github.com/ambv/black for code formatting could help 👍 

My setup.cfg
```cfg
[flake8]
import-order-style = google
application-import-names = core # Put the app name here
exclude = .git,*migrations*,__pycache__,*.pyc
max-line-length = 88
select = C,E,F,W,B,B3,B9 
ignore = W503, E501, E203
```
some dev dependencies I have.

```toml
[dev-packages]
"flake8" = "*"
"flake8-import-order" = "*"
"flake8-bugbear" = "*"
pytest = "*"
the = "*"
black = "*"
isort = "*"
pydocstyle = "*"
```

josephmancuso:
Nvm I think that last fix fixed it. Was so annoying because it was an import inside a method lol

PR #367, opened 2018-09-25
change username to email

vaibhavmule:


vaibhavmule:
Anyways user going to change controller when default it or username, user should also template too.

vaibhavmule:
@josephmancuso Should I close it as this PR will be addressed by @bjorntheart  in #369 

Or Are we thinking for having a feature where the user can username based or email based login?

PR #375, opened 2018-09-28
Removed the ability to delete compiled sass file after test is ran.

hammacktony:
Removed `os.remove(os.path.join(os.getcwd(), 'storage/compiled/style.css'))`.

When you run pytest more than once, `test_compile_sass.py` fails. No need to remove it, since you know that it exists and that is the main function of the test.


vaibhavmule:
Sorry for typos, I'm using mobile, I meant for cors middleware file.

vaibhavmule:
@hammacktony Good Catch. I have gone through that but didn't notice that.

vaibhavmule:
This has to be reverted as 

Tests requires:

1. creates the file
2. checks if it exists
3. removes the file (This is part is important as next time you run the tests, it work as expected.)

Revert: https://github.com/MasoniteFramework/core/pull/382

vaibhavmule:
@hammacktony 

If the tests are failing, make sure you have `libsass` installed.

PR #385, opened 2018-09-28
added azure and rackspace upload drivers

josephmancuso:
This PR adds two new upload drivers, Rackspace and Azure


vaibhavmule:
@josephmancuso How do you test these drivers?

bjorntheart:
This should probably move to the #internals channel in Slack

When I think filenames in the context of uploading files, I see users wanting these options

1. upload the file with the original filename -`testme.png`
2. upload the file with a totally random string generated filename - `3bcc367a3488e113dca68b67e5fa262fe4fd2df48b1b72fd3292b30358911aab.png` in this case I **sha256** the filename
3. combination of 1 and 2 - `testme-3bcc367a3488e113dca68b67e5fa262fe4fd2df48b1b72fd3292b30358911aab.png`

@josephmancuso Do you think this is something useful we can implement for the file upload drivers?

josephmancuso:
I completely agree. I honestly even think the drivers we have currently are really primitive. They can currently only upload cgi.StorageFile type objects. I'll deff need to sit down and read some documentation on how we can really make these awesome.

I'm not even really a fan of the `store_prepend` method. Just looks ugly.

But yes I agree 100%

josephmancuso:
as for the 25 and 15, just me forgetting which one I used before :) lol

josephmancuso:
@vaibhavmule as for testing 🤔 if we test by uploading something, you're going to be using a crap ton of IO via the services we are using which will incurr a whole bunch of charges so what i'll do is set a envrionment variable and test them "manually" by only running those set of tests if an environment variable tells it to do so.

josephmancuso:
@bjorntheart I missed your question earily. So the rackspace driver needs to upload files (which need to be saved to disk first) but the azure driver can upload objects directly.

josephmancuso:
added these to Masonite Collective instead https://github.com/bjorntheart/masonite-collective/pull/1

Issue #370, opened 2018-09-26
integrate Masonite events into core 


josephmancuso:
We should be able to add several events that can be listened to throughout the codebase. Let's make Masonite Events a dependency for core (and throw it in the container via the Events Service Provider) and that's it for this issue. Once that is done we will open a new issue for adding specific events

josephmancuso:
it should be as simple as following the installation steps in the Masonite Events docs in the official docs website

rickwest:
Can I take this @josephmancuso? Would love to start contributing and this seems like a nice way in!

josephmancuso:
sure! Let me know if you have any questions. Either drop them here and @ me or in the slack channel is fine :)

clsource:
I think it would be a good idea that the core contains the files of event system inside the repo and not as a dependency. Since it would be a core functionality it makes sense. Also it seems that the events repo Does not have many stars that would affect such merge

PR #386, opened 2018-09-28
Integrate masonite-events into core

rickwest:
References #370. Have added the provider to providers list as per docs. Do I also need to add masonite-events as a requirement in setup.py. Forgive my ignorance, I'm not very familiar with dependancy management in Python.


josephmancuso:
Yes

vaibhavmule:
@rickwest Do not worry about Travis build is failing, 

I can see that https://github.com/MasoniteFramework/events/releases/tag/v1.0.1 is not released on [pypi](https://pypi.org/project/masonite-events/#history).

This should merge after, https://github.com/MasoniteFramework/core/issues/387

rickwest:
@vaibhavmule Thank you 🙂

josephmancuso:
fixed #387 and rerunning tests

josephmancuso:
@clsource the problem I have with merging things is that although we are going to be using it in core pretty extensively, I can foresee Masonite Events as being much bigger than it currently is which will really bloat core. I'd like core to be ... core ... and then we sort of develop packages and even other developers develop packages and then we can just depend on them.

I just don't want to bloat core with a crap ton of packages. It might be small now but it could get massive once we start thinking of useful features

PR #389, opened 2018-09-30
added ability to merge containers

josephmancuso:
This might be really dumb but I made a way to merge several containers together. I've been reading a lot about software architecture and how we can build huge systems in maintainable ways and read https://github.com/Mahmoudz/Porto as reference.

The idea behind merging containers is that you can have separate standalone applications that are very similiar in nature (like 2 Masonite projects) that can be combined together to form 1 application. See tests for examples on what it can do


theodesp:
The tricky part is to compete with the microservices approach where everything is low-coupled as it's not clear how is this going to improve that.

josephmancuso:
@MasoniteFramework/reviewers what do we think of this?

josephmancuso:
closing. This doesn't seem very useful I don't think

Issue #371, opened 2018-09-27
Add Log Mail Driver for test purposes.


mitchdennett:
It could be nice to have another mail driver that just prints out the email string/html to a log or console especially when developing/test.

vaibhavmule:
I agree with as it will help a lot i development

josephmancuso:
oh that'd be really cool

mitchdennett:
I'll take this if no one else has started

mitchdennett:
Yeah just the terminal for now.  We could probably have an option in Mail.py to enter a file and then save the log to a file instead.

See below:

![screen shot 2018-09-30 at 9 26 45 pm](https://user-images.githubusercontent.com/16268619/46265762-b5bf9d00-c4f7-11e8-9fc4-21338bd264cf.png)

josephmancuso:
this is being worked on in the PR listed above ☝️

PR #390, opened 2018-10-01
Added MailLogDriver and MailTerminalDriver

mitchdennett:
Closes Issue #371


mitchdennett:
Hmm yeah it would probably be more clear to the user what is happening if we had 2 different ones.
I'll sort that out today. Should only be a small change.

josephmancuso:
freakin awesome 🏅

mitchdennett:
Oh yeah I forgot to ask where we should store it ;)

PR #391, opened 2018-10-01
add better naming for api controller methods

vaibhavmule:
reference:

Phoenix Framework
```
user_path  GET     /users           HelloWeb.UserController :index
user_path  GET     /users/:id/edit  HelloWeb.UserController :edit
user_path  GET     /users/new       HelloWeb.UserController :new
user_path  GET     /users/:id       HelloWeb.UserController :show
user_path  POST    /users           HelloWeb.UserController :create
user_path  PATCH   /users/:id       HelloWeb.UserController :update
           PUT     /users/:id       HelloWeb.UserController :update
user_path  DELETE  /users/:id       HelloWeb.UserController :delete
```

Rails:
```
GET	/photos	photos#index	display a list of all photos
GET	/photos/new	photos#new	return an HTML form for creating a new photo
POST	/photos	photos#create	create a new photo
GET	/photos/:id	photos#show	display a specific photo
GET	/photos/:id/edit	photos#edit	return an HTML form for editing a photo
PATCH/PUT	/photos/:id	photos#update	update a specific photo
DELETE	/photos/:id	photos#destroy	delete a specific photo
```


clsource:
Maybe different templates could be used to emulate masonite, phoenix and rails ways.

josephmancuso:
I keep looking at this PR but what is "better" about these?

vaibhavmule:
I'm not sure about the better, but the developer coming from MVC based framework background might find it confsuing,
When you say in `show` is showing all resources and `index` is show single resource.

We can keep `store` but then I wanted to follow CRUD naming. so changed `store` to `create` and `destroy` to `delete`

josephmancuso:
🤔 hmm ok I think i'm on board now but I'm not a huge fan of `new` and `create`. They would sound the same to me

vaibhavmule:
Yeah, you are right. They do sound similar and not convey what it does, create I can understand, but yeah, `create` vs `new` is confusing to me too.

josephmancuso:
I'm gonna close this until we are in more of an agreement on what to call the methods

PR #395, opened 2018-10-01
Per convention, using _ as a general purpose throwaway variable name

vaibhavmule:


clsource:
seems ok to me

josephmancuso:
I THINK I tried it but pylint kept throwing a red squiggly on this but I'll double check when I merge this in. Can you confirm pylint does not throw any red squiggly?

vaibhavmule:
it does not,
check,
screenshot 2018-10-01 at 6 06 51 pm

PR #402, opened 2018-10-04
Add amqp queue driver

josephmancuso:
This PR adds support for real message brokers for queues. Currently we just have an async driver which sends jobs into a seperate thread which works pretty well actually but this PR introduces support for any `amqp` standardized message broker like Rabbitmq.

Once this PR gets merged the way to use it will be:

1. install RabbitMQ on your machine (and production machine)
2. change the driver to `amqp` and add the required credentials:

```python
DRIVER = 'amqp'
...
DRIVERS = {
    'amqp': {
        'username': 'guest',
        'password': 'guest',
        'host': 'localhost',
        'port': '5672',
        'channel': 'default',
    }
}
```

3. Start RabbitMQ. Something like:

```bash
$ rabbitmq-server

  ##  ##
  ##  ##      RabbitMQ 3.7.8. Copyright (C) 2007-2018 Pivotal Software, Inc.
  ##########  Licensed under the MPL.  See http://www.rabbitmq.com/
  ######  ##
  ##########  Logs: /usr/local/var/log/rabbitmq/rabbit@localhost.log
                    /usr/local/var/log/rabbitmq/rabbit@localhost_upgrade.log

              Starting broker...
 completed with 6 plugins.
```

which starts the server.

4. Then start the Masonite worker in a separate terminal:

```bash
$ craft queue:work
```

5. Pass jobs into the queue like normal:

```python
from app.jobs import SomeJob
...
def show(self, Queue):
    # do your normal logic
    Queue.push(SomeJob)
```

and you should see them in the worker terminal window:

screen shot 2018-10-04 at 12 03 31 am

That's it!

### DIFF
--- a/masonite/commands/QueueWorkCommand.py
+++ b/masonite/commands/QueueWorkCommand.py
@@ -1,5 +1,7 @@
 """ A QueueWorkCommand Command """
+
 import pickle
+import inspect
 
 from cleo import Command
 
@@ -10,7 +12,9 @@ from masonite.exceptions import DriverLibraryNotFound
 
 def callback(ch, method, properties, body):
     from wsgi import container
-    job = container.resolve(pickle.loads(body))
+    job = pickle.loads(body)
+    if inspect.isclass(job):
+        job = container.resolve(job)
     job.handle()
     ch.basic_ack(delivery_tag=method.delivery_tag)
 
--- a/masonite/commands/QueueWorkCommand.py
+++ b/masonite/commands/QueueWorkCommand.py
@@ -25,9 +25,7 @@ class QueueWorkCommand(Command):
 
     queue:work
         {--c|channel=default : The channel to listen on the queue}
-        {--e|exchange=default : The channel to listen on the queue}
         {--f|fair : Send jobs to queues that have no jobs instead of randomly selecting a queue}
-        {--d|durable : Send jobs to queues that have no jobs instead of randomly selecting a queue}
     """
 
     def handle(self):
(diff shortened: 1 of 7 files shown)


bjorntheart:
I'm very excited about this one @josephmancuso. Great work

josephmancuso:
we will need to find a way to pass in variables to the job. I am thinking we do it like Laravel and just instantiate the object instead of passing the class:

```python
from app.jobs import SomeJob
...
def show(self, Queue):
    # do your normal logic
    Queue.push(SomeJob(variable1, variable2))
```

But if this is going into the minor we also need to support this for `async` driver. This MAY need some workaround for 2.0 that won't be required for 2.1. 

I want to make sure this goes into the next minor so @aisola and I can use it in some current projects we have.

josephmancuso:
@MasoniteFramework/reviewers Without any objections, I'm going to merge this in tonight.

josephmancuso:
The latest commit also adds the ability to do 2 things:

1. You can add the class like normal:

```python
from app.jobs import SayHello

def show(self, Queue):
    Queue.push(SayHello)
```

This will resolve the constructor

Or you can pass now pass in an object and instantiate it which is useful for passing in variables from the container:

```python
from app.jobs import SayHello

def show(self, Queue):
    Queue.push(SayHello('variable1', 'variable2'))
```

Instantiating the class will NOT resolve the constructor. So it is currently either have the constructor resolved or pass in data from controller (which can include any dependencies you need anyway so not THAT big of a deal.)

PR #409, opened 2018-10-04
Fixes #408 Removed api.py and references to it

cmpickle:

PR #410, opened 2018-10-04
Update AppProvider.py

CadenG150:


josephmancuso:
looks like https://github.com/MasoniteFramework/core/pull/409 got this issue Already :)

PR #411, opened 2018-10-05
Add swap to container

josephmancuso:
This PR adds the ability to "substitute" whatever objects are resolving from the container. For example if we want to specify one object but actually resolve a different object. 

We can do so by specifying an object we want substituted and then specifying a callback:

```python
app.substitute(SubstituteThis, self._substitute)
```

self.substitute is function or callable object:

```python
    def _substitute(self, method, container):
        return 'test'
```

now anytime we resolve `SubstituteThis` it will run this method and return 'test' instead:

```python
    def _test_substitute(self, test: SubstituteThis):
        return test # =='test'
```


josephmancuso:
changed substitute to swap because I was having a hard time always having to spell substitute

josephmancuso:
added ability to swap just an object. So originally we passed in some callable like a function which receives 2 parameters but if we dont pass in a callable, it will return that value we supplied:

```python
app.substitute(SubstituteThis, 'test')
```

now anytime we resolve SubstituteThis it will run this method and return 'test' instead:

```python
    def _test_substitute(self, test: SubstituteThis):
        return test # =='test'
```

PR #412, opened 2018-10-05
Fixes #408 Removed api.py and references to it

### LINKED ISSUE #408: remove routes/api.py
There is no need for this file and the new api package will not need this so this can be removed. This involved deleting it from the masoniteframework/masonite repo as well as removing any references from this repo. 

I think the only reference from this repo is in the `AppProvider`

cmpickle:


josephmancuso:
Thank you very much 😄

PR #423, opened 2018-10-08
extends BaseScaffoldCommand for ViewCommand

### LINKED ISSUE #422: Craft view auth/home create directory if it doesn't exist
`craft view auth/home` should create directory if it the `auth` directory does not exists.

This way I reduce my time creating directory before creating any view.

vaibhavmule:
Closes #422


vaibhavmule:
I forgot that I have to based of from master branch.

I will do another PR for which will be based of master.

I will keep this PR as it is because this how it should be in 2.1

PR #431, opened 2018-10-08
Extend Method Removed from View Class

### LINKED ISSUE #426: remove extend method from View class
https://github.com/MasoniteFramework/core/blob/master/masonite/view.py#L133

Kush22:
Closes #426

### DIFF
--- a/masonite/view.py
+++ b/masonite/view.py
@@ -130,7 +130,6 @@ class View:
 
         return self
 
-
     def share(self, dictionary):
         """Shares data to all templates.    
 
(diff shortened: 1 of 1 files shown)


masonite/view.py
```diff
@@ -130,11 +130,6 @@ def composer(self, composer_name, dictionary):
 
         return self
 
-    def extend(self):
-        """Extend class.
-        """
-
-        pass
```
vaibhavmule: please remove this line too as Pep8 suggests 1 line space between methods in the class.

vaibhavmule:
@Kush22 Good work,

Thanks for contributing.

PR #434, opened 2018-10-08
Redis Cache Driver

mitchdennett:
This PR is for implementing Redis as a Cache Driver. There are some (a lot more) configuration settings we could allow the user to set. Not sure how many we want to add straight away or just see how this works and add them as people request them?

### DIFF
--- a/config/cache.py
+++ b/config/cache.py
@@ -2,10 +2,17 @@
 Cache configuration
 """
 
+import os
+
 DRIVER = 'disk'
 
 DRIVERS = {
     'disk': {
         'location': 'bootstrap/cache'
+    },
+    'redis': {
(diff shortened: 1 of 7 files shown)


josephmancuso:
This is awesome!

josephmancuso:
Just had to make a slight modification to the tests. Now only the test runs when the `REDIS_CACHE_DRIVER` environment variable is true. This way I can run the tests on my or your computer when we need to instead of needing a Redis server every time we run tests or needing to set one up for Travis

josephmancuso:
Might be worth looking at how I modified the tests though: https://github.com/MasoniteFramework/core/blob/master/tests/test_cache.py#L25-L30

mitchdennett:
Good catch about the tests. I didn't think about that.

PR #435, opened 2018-10-10
removed messy header handling

josephmancuso:
This PR cleans up some code a bit and prevents the need to handle an arbitrary headers key in the container


vaibhavmule:
That's a good clean up!

masonite/exception_handler.py
```diff
@@ -74,3 +75,6 @@ def _render(self):
                                                }
                                                ).rendered_template
         self._app.bind('Response', rendered_view)
+        request.header('Content-Type', str(len(rendered_view)))
+        self._app.bind('Headers', request.get_headers())
+        request.reset_headers()
```
vaibhavmule: Why do we need to reset header and also why do we need request.get_headers, when you already set headers by request.header?

masonite/providers/StatusCodeProvider.py
```diff
@@ -30,20 +31,18 @@ def register(self):
...
-                    'code': self.app.make('StatusCode')
+                    'code': request.get_status_code()
                 }).rendered_template
-            Headers = [
-                ("Content-Length", str(len(rendered_view)))
-            ]
-            self.app.bind('Response', rendered_view)
 
-            self.app.bind('Headers', Headers)
+            self.app.make('Request').header('Content-Length', str(len(rendered_view)))
```
vaibhavmule: if I'm not wrong this should be `request.header`?

PR #436, opened 2018-10-10
remove setting `Content-Length` for 404, 500, 503

vaibhavmule:
This code sets Content-Length Twice

screenshot 2018-10-10 at 6 24 57 pm


josephmancuso:
We're going to need to keep this in. but we have to figure out how to prevent setting headers twice

josephmancuso:
the reason we need to keep this in is that the above template has a different content length and needs to override any content length already set

josephmancuso:
actually I will experiment with this. I was going to close because there is an open PR that fixes this going into develop but this may need a hotfix

PR #437, opened 2018-10-11
Add env

josephmancuso:
This PR adds a new `env` function which helps convert environment types to their respective boolean or int values.

we are currently doing this:

```python
import os
...
DATABASES = {
    'default': {
        'driver': os.environ.get('DB_DRIVER', 'default'),
        'host': os.environ.get('DB_HOST', 'default'),
        'port': os.environ.get('DB_PORT', 'default'),
    }
}
```

Now we can do this:

```python
from masonite import env
...
DATABASES = {
    'default': {
        'driver': env('DB_DRIVER', 'default'),
        'host': env('DB_HOST', 'default'),
        'port': env('DB_PORT', 'default'),
    }
}
```

Not only does this shorten things down a bit but it ALSO will type cast as well. So an environment variable of `'3306'` (string) will actually convert to `3306` (int) 

Also an environment variable of `'True'` or `'true'` will convert to the actual boolean of `True`

### DIFF
--- a/masonite/environment.py
+++ b/masonite/environment.py
@@ -45,3 +45,14 @@ class LoadEnvironment:
 
         env_path = str(Path('.') / '.env.{}'.format(env))
         load_dotenv(dotenv_path=env_path, override=override)
+
+def env(value, default=None):
+    env_var = os.getenv(value, default)
+    if env_var.isnumeric():
+        return int(env_var)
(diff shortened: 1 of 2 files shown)


josephmancuso:
This actually also solves the problem of `'False'` actually being `True` because it is a string

PR #447, opened 2018-10-16
Show the container name instead of `This`

vaibhavmule:
Before:

```
masonite.exceptions.ContainerError: This container is not set to resolve parameters. You can set this in the container constructor using the 'resolve_parameters=True' keyword argument.
```

After:
```
masonite.exceptions.ContainerError: MailConfig container is not set to resolve parameters. You can set this in the container constructor using the 'resolve_parameters=True' keyword argument.
```


josephmancuso:
well mailconfig is not a container it is a dependency that the container is trying to resolve

PR #453, opened 2018-10-18
Reset password

### LINKED ISSUE #262: Make password reset functionality along with auth command

vaibhavmule:
Closes #262 

Inspired by https://github.com/josephmancuso/gbaleague-masonite2

### DIFF
--- a/masonite/snippets/auth/controllers/PasswordController.py
+++ b/masonite/snippets/auth/controllers/PasswordController.py
@@ -1,0 +1,49 @@
+"""A PasswordController Module."""
+
+import uuid
+
+from masonite import env
+from masonite import Mail
+from masonite import Session
+from masonite.auth import Auth
+from masonite.helpers import password as bcrypt_password
+from masonite.request import Request
+from masonite.view import View
+
+from app.User import User
+
+
+class PasswordController:
+    """Password Controller."""
+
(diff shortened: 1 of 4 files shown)


masonite/snippets/auth/controllers/PasswordController.py
```diff
@@ -0,0 +1,49 @@
...
+
+from masonite import env
+from masonite import Mail
+from masonite import Session
+from masonite.auth import Auth
+from masonite.helpers import password as bcrypt_password
+from masonite.request import Request
+from masonite.view import View
+
+from app.User import User
```
josephmancuso: Don't import the user model directly. it could be moved around. Instead you should import the auth config model. Like:

```
from config.auth import AUTH

AUTH['model] # == User model or whatever model they choose to be the authentication model
```

masonite/snippets/auth/controllers/PasswordController.py
```diff
@@ -0,0 +1,49 @@
+"""A PasswordController Module."""
+
+import uuid
+
+from masonite import env
+from masonite import Mail
+from masonite import Session
```
josephmancuso: also condense these down:

```
from masonite import env, Mail, Session
```

vaibhavmule:
@josephmancuso now check.

PR #457, opened 2018-10-30
Add pipenv support

hammacktony:
Enabled Pipenv Support.


josephmancuso:
let's close these down. If we are going to go with pipenv let's try to start fresh again

PR #465, opened 2018-11-03
Update Orator

hammacktony:
Updated `orator` to 0.9.8. Vast bug improvements that can benefit uses.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ MarkupSafe==1.0
 more-itertools==4.3.0
 msgpack-python==0.5.6
 ndg-httpsclient==0.5.1
-orator==0.9.7
+orator==0.9.8
 pastel==0.1.0
 pendulum==1.4.4
 pluggy==0.7.1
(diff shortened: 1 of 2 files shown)


hammacktony:
If you check the error log for this PR, Python 3.4 raises an interesting error. It is the classic chicken and the egg scenario. Beta 2.1 requires `orator==0.9.7`, but that is what I am working on at the moment. 

The 2.1 branch needs to be updated first. 

Otherwise, this PR is ready to go. It works for 3.5, 3.6, and 3.7. 3.4 is being stubborn.

josephmancuso:
that is super bizarre 🤔

hammacktony:
I did the `>=0.9.7`. For most systems, it will default to `0.9.8`

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

hammacktony:
Yeah. It’s weird. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

josephmancuso:
i have no idea why that release broke 3.4. That is not cool. I really like your fix though. Makes a lot of sense

PR #470, opened 2018-11-06
fixed requests setting double headers

josephmancuso:
This PR fixes an issue that resulted in mutliple of the same header being sent


masonite/request.py
```diff
@@ -358,7 +356,14 @@ def get_headers(self):
             dict -- Dictionary of all headers.
```
vaibhavmule:
```suggestion
            list -- Array of all headers.
```

PR #472, opened 2018-11-08
New reloading development server

aisola:
I was chatting with @josephmancuso and found several issues regarding the quality of waitress + our old code reloading system. We found that the development experience could be greatly improved by changing it out with this instead.

The improvements are:
- No more waitress bugs, we now use a slightly modified wsgiref-based server for development. This won't matter to most people but it will matter if you use frameworks on the front end or if you accept a lot of files.
- The code reloading is much faster. We're letting the operating system do its job now.

I've tested this on both Mac OSX and Ubuntu. I would appreciate if somebody could test this on Windows. It wouldn't hurt to have somebody confirm that it works on both Mac and Linux again too.


masonite/commands/ServeCommand.py
```diff
@@ -20,20 +14,24 @@ class ServeCommand(Command):
...
         if self.option('reload'):
-            self._run_with_reloader(extra_files=[".env"])
+            # start_reloader will only return in a monitored subprocess
+            reloader = hupper.start_reloader('masonite.commands._devserver.run', worker_args=[
+                # worker args are pickled and then passed to the new process
+                self.option("host"), self.option("port"), "wsgi:application",
+            ])
+
+            # monitor the env file too
+            reloader.watch_files(['.env'])
```
josephmancuso: This line of code is never reached

josephmancuso:
Made a slight modification. There was an issue where a file not found was being thrown. Apparently i think the loader is local to the commands directory? prepending the cwd to the file did the trick.

screen shot 2018-11-13 at 11 12 20 pm

masonite/commands/ServeCommand.py
```diff
@@ -20,20 +18,47 @@ class ServeCommand(Command):
...
+    def _run_reloader(self, reloader, extra_files=[]):
+        reloader._capture_signals()
+        reloader._start_monitor()
+        for blob in extra_files:
+            reloader.monitor.add_path(os.path.join(os.getcwd(), blob))
+        try:
+            while True:
+                if not reloader._run_worker():
+                    reloader._wait_for_changes()
+                time.sleep(reloader.reload_interval)
```
josephmancuso: I think i understand what this code is doing: giving time for the PID to be killed before starting a new one?

but heres a crazy idea: if we comment this line of code out, the reloader speed increases by like 23x. Thats more than an order of magnitude faster. See the screen shots attached. With it there the reloader is 92 microseconds but with it commented out it is around 4-5 micro seconds

screen shot 2018-11-13 at 11 41 02 pm
screen shot 2018-11-13 at 11 41 27 pm

josephmancuso: Now what I THINK i have noticed though is that it does indeed not kill a process very very occasionally. and even when it doesn't LOOK like it kills the process, i'm not sure it doesn't. Take these snippets for example. If i save and edit SUPER fast. Like 3 saves per seconds going crazy in order to mock the starting and killing of proccesses very fast I can get the process to start but never kill in the terminal. 

BUT i can never get more then 2 processes in activity monitor. So i'm not sure it doesn't kill it. This is with the sleep line of code commented out.

screen shot 2018-11-13 at 11 52 37 pm
screen shot 2018-11-13 at 11 51 09 pm

josephmancuso: actually relooking at it it does actually kill it right below it. I didn't see that. I honestly think we can just uncomment that sleep code and speed this up by ~25x

PR #479, opened 2018-11-14
added more classs directly to Masonite namespace

josephmancuso:
added a better import. Now we can alternatively do this:

```python


def show(self ..):
    ..
```

and do this instead:

```python
from masonite import Request, View

def show(self ..):
    ..
```

### DIFF
--- a/masonite/__init__.py
+++ b/masonite/__init__.py
@@ -5,3 +5,6 @@ from .managers.QueueManager import Queue
 from .managers.SessionManager import Session
 from .managers.UploadManager import Upload
 from masonite.environment import env
+from .request import Request
+from .view import View
+from .provider import ServiceProvider


masonite/providers/AppProvider.py
```diff
@@ -27,10 +28,10 @@ class AppProvider(ServiceProvider):
     def register(self):
         self.app.bind('HookHandler', Hook(self.app))
         self.app.bind('WebRoutes', flatten_routes(web.ROUTES))
-        self.app.bind('Response', None)
         self.app.bind('Storage', storage)
         self.app.bind('Route', Route())
         self.app.bind('Request', Request())
+        self.app.simple(Response(self.app))
```
vaibhavmule: @josephmancuso Did you experience this error?

```
WSGI
Application
Container
ProvidersConfig
Providers
WSGIProviders
HookHandler
WebRoutes
Storage
Route
Request
<class 'masonite.response.Response'>

Traceback (most recent call last):
  File "/Users/vaibhavmule/.local/share/virtualenvs/pythonpips-EagQ3vWO/bin/craft", line 7, in <module>
    from masonite_cli.application import application
  File "/Users/vaibhavmule/.local/share/virtualenvs/pythonpips-EagQ3vWO/lib/python3.7/site-packages/masonite_cli/application.py", line 44, in <module>
    if key.endswith('Command'):
AttributeError: type object 'Response' has no attribute 'endswith'
```

PR #489, opened 2018-11-28
[Develop] Modify header method in request to return empty string 

### LINKED ISSUE #488: Getting a header should return empty string instead of None
Currently when we do something like:

```python
request.header('Content-Type')
```

it will return None if one doesn't exist but this will cause issues when people use it as an interator:

```python
if 'text/html' in request.header('Content-Type')
```

This will throw an exception because` None` is not an iterator.

## This should be based off the develop branch

Te7a-Houdini:
this **PR** , resolves #488 by changing ```None``` to ```Empty String```

### DIFF
--- a/masonite/request.py
+++ b/masonite/request.py
@@ -356,7 +356,7 @@ class Request(Extendable):
             if key in self.environ:
                 return self.environ[key]
             else:
-                return None
+                return ''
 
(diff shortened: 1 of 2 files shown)


josephmancuso:
Looks good to me :) thanks for contributing 👍 ❤️

PR #504, opened 2018-12-11
Fix container error

### LINKED ISSUE #501: Need to revisit: ContainerError > Either Tried resolving the too many dependencies
That error is thrown a lot because of an if statement in the resolver. This needs to be fixed and either throw one or the other

josephmancuso:
This completely removed the beginning text which seemed to be always incorrect.

I think the "tried resolving too many dependencies" thing was something i encountered mid development. I can't seem to replicate that use case. I remember coding and then refactoring a bit so maybe that is no longer the case anymore.

Closes #501


masonite/app.py
```diff
@@ -223,6 +223,9 @@ def _find_annotated_parameter(self, parameter):
         raise ContainerError(
             'The dependency with the {0} annotation could not be resolved by the container'.format(parameter))
 
+    def get_parameters(self, obj):
```
vaibhavmule: Where this will get used?

josephmancuso: just updated

josephmancuso: :)

vaibhavmule: oh lol, I jumped in the middle of your work.

josephmancuso: np lol. I was actually done but we caught that at the exact same time and I updated PR

PR #507, opened 2018-12-19
Create SecureHeadersMiddleware.py

### LINKED ISSUE #502: Add secure headers middleware
Make a middleware available that you are able to have secure headers available to your project by simply adding it to the middleware stack:

```
from masonite.middleware import SecureHeadersMiddleware

HTTP_MIDDLEWARE = [
    ...
    SecureHeadersMiddleware,
]
```

vaibhavmule:
Closes #502

### DIFF
--- a/masonite/middleware/SecureHeadersMiddleware.py
+++ b/masonite/middleware/SecureHeadersMiddleware.py
@@ -38,5 +38,5 @@ class SecureHeadersMiddleware:
             self.headers.update(middleware.SECURE_HEADERS)
         except AttributeError:
             pass
-            
+
         self.request.header(self.headers)
--- a/masonite/middleware/SecureHeadersMiddleware.py
+++ b/masonite/middleware/SecureHeadersMiddleware.py
@@ -1,0 +1,39 @@
+"""SecureHeaders Middleware."""
+
+from masonite.request import Request
+
+from config import middleware
+from config.application import DEBUG
+
+
+class SecureHeadersMiddleware:
+    """SecureHeaders Middleware."""
+
+    def __init__(self, request: Request):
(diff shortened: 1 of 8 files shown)


masonite/middleware/SecureHeadersMiddleware.py
```diff
@@ -0,0 +1,39 @@
...
+            'Strict-Transport-Security': 'max-age=63072000; includeSubdomains',
+            'X-Frame-Options': 'SAMEORIGIN',
+            'X-XSS-Protection': '1; mode=block',
+            'X-Content-Type-Options': 'nosniff',
+            'Referrer-Policy': 'no-referrer, strict-origin-when-cross-origin',
+            'Cache-control': 'no-cache, no-store, must-revalidate',
+            'Pragma': 'no-cache',
+        }
+        headers.update(middleware.SECURE_HEADERS)
+        if not DEBUG:
```
clsource: Why this headers would be added only when not in DEBUG mode?.
Just need more clarification about this 👍

josephmancuso:
Also need to add tests. Since you made a branch from this repo I can add them tonight if I get around to it. Test really would be just passing in a request object and then asserting some of the headers were set correctly.

masonite/middleware/SecureHeadersMiddleware.py
vaibhavmule: Actually, I have mixed feeling about it. 

I only needed it in a production environment, but I think we can remove `not Debug` and give the ability to developer comment out SecureHeadersMiddleware in `config/middleware.py`

vaibhavmule:
@josephmancuso I wasn't  sure about branch that's why I choose 2.1

PR #508, opened 2018-12-21
test cases for secure headers

vaibhavmule:


vaibhavmule:
Sorry about this PR as I wasn't aware that I created a branch in core rather than vaibhavmule/core

josephmancuso:
honestly it's better if you create it in this core because I can tweak things as needed. I haven't found out how to tweak PR's from other people branches without getting permission first

josephmancuso:
but this is fine, don't move it now

PR #521, opened 2019-01-09
Add ability to login using multiple columns

### LINKED ISSUE #516: Add ability to make multiple matches for __auth__ attribute on user model
Currently we are able to match either a username or email on login but we may need to sometimes match both.

currently we have something like:

```python
class User(Model):
    """User Model
    """

    __fillable__ = ['name', 'email', 'password']

    __auth__ = 'email'
```

to match the `email` but we should be able to match several fields in an `or` clause like this:

```python
class User(Model):
    """User Model
    """

    __fillable__ = ['name', 'email', 'password']

    __auth__ = ['email', 'name']
```

this will either find matches for the email or name field in addition to the password supplied on authentication

## branch off of master as this is going into the next minor version as a feature

josephmancuso:
This PR adds the ability to login using multiple columns. For example we may have a form that we can login via both the `username` OR the `email`.

In order to login using either or column we can now set the user model `__auth__` to a list:

Instead of:

```python
class User(Model):
    __auth__ = 'email'
```

we can do:

```python
class User(Model):
    __auth__ = ['name', 'email']
```

Additionally this PR also adds the ability to specify a `__password__` attribute if your `password` column is not called `password` like this:

```python
class User(Model):
    __password__ = 'user_password'
```

Closes #516


masonite/auth/Auth.py
```diff
@@ -63,9 +63,23 @@ def login(self, name, password):
...
         try:
-            model = self.auth_model.where(auth_column, name).first()
+            password_column = self.auth_model.password if hasattr(self.auth_model, 'password') else self.auth_model.__password__
+        except AttributeError as e:
+            raise AttributeError('Your model does not have a password column or a designated __password__ attribute. Set the __password__ attribute to the name of your password column.') from e
 
-            if model and bcrypt.checkpw(bytes(password, 'utf-8'), bytes(model.password, 'utf-8')):
+        try:
+            # Try to login multiple or statements if given an auth list
+            if isinstance(auth_column, list):
```
vaibhavmule: so the list can more than 2 fields to authenticate the user?

['username', 'email', 'mob_number']

josephmancuso: yes. It will take the first one and do a `where`, and then do a `or_where` for each other element in the list.

vaibhavmule: wow, this is good feature. LGTM

PR #522, opened 2019-01-09
spring cleaning

vaibhavmule:
tests are successfully except for `tests/testing/test_unit_test.py` and `tests/test_upload_manager.py`

@josephmancuso can you have a look?


config/middleware.py
```diff
@@ -1,51 +1,41 @@
...
+"""HTTP Middleware
+HTTP middleware is middleware that will be ran on every request. Middleware
+is only ran when a HTTP call is successful (a 200 response). This list
+should contain a simple aggregate of middleware classes.
 """
 
 HTTP_MIDDLEWARE = [
     LoadUserMiddleware,
     # todo
     # CsrfMiddleware,
```
vaibhavmule: @josephmancuso FYI. I have kept it as todo.

config/database.py
```diff
@@ -47,3 +60,16 @@
...
+logger.setLevel(logging.DEBUG )
+
+formatter = logging.Formatter(
+    'It took %(elapsed_time)sms to execute the query %(query)s'
+)
+
+handler = logging.StreamHandler()
+handler.setFormatter(formatter)
+
+logger.addHandler(handler)
```
josephmancuso: What is this addition?

config/middleware.py
josephmancuso: to uncomment it?

config/storage.py
```diff
@@ -1,84 +1,79 @@
...
     },
     's3': {
-        'client': os.getenv('S3_CLIENT', 'AxJz...'),
-        'secret': os.getenv('S3_SECRET', 'HkZj...'),
-        'bucket': os.getenv('S3_BUCKET', 's3bucket'),
-        'location': 'http://s3.amazon.com/bucket',
-    }
+        'client': env('S3_CLIENT', 'AxJz...'),
+        'secret': env('S3_SECRET', 'HkZj...'),
+        'bucket': env('S3_BUCKET', 's3bucket'),
```
josephmancuso: The reason the tests are failing is because you are missing the `location` here

config/middleware.py
vaibhavmule: I mean I kept it as it is. it commented out when I doing spring cleaning.

config/database.py
vaibhavmule: I'm not sure but they are in masonite repo.
https://github.com/MasoniteFramework/masonite/blob/8f7a289e1e918b58c1115d669bea8fbe9b8a8a68/config/database.py#L55

josephmancuso: 🤔 i'm not sure why i added these .... lol

config/storage.py
vaibhavmule: thanks for pointing out

config/database.py
vaibhavmule: is it related to logger driver?

josephmancuso: no it's suppose to show the queries in the terminal (so you can optimize) but i thought thats what the `DB_LOG` was doing .. lol

josephmancuso: not sure why i added this extra code

vaibhavmule: we can keep it for now and remove it later.

vaibhavmule:
@josephmancuso Now it is failing at session, did i miss something?
https://travis-ci.org/MasoniteFramework/core/jobs/477381795

config/session.py
```diff
@@ -1,18 +1,13 @@
...
+"""Session Driver
+Sessions are able to be linked to an individual user and carry data from
+request to request. The memory driver will store all the session data
+inside memory which will delete when the server stops running.
+
+Supported: 'memory', 'cookie'
 """
 
-DRIVER = os.getenv('SESSION_DRIVER', 'memory')
+DRIVER = env('SESSION_DRIVER', 'cookie')
```
josephmancuso: possibly this change from cookie to memory?

josephmancuso: this shouldn't break it tho but i wonder if this is breaking it

josephmancuso: does it work locally?

vaibhavmule: No it does not, memory works.

josephmancuso: hmm i'm gonna need to bring that in later and make sure nothing is wrong with that

config/database.py
vaibhavmule: ?

PR #527, opened 2019-01-18
Added ability on login and login_by_id in Auth to

ralliart2004:
set a cookie expiration via the request.cookie
method via environment variable. Added ability
to do cookie only authentication via environment
variable of the logged in user if the cookie exists.
Default for COOKIE_EXPIRE_TIME empty string
Default for COOKIE_ONLY_AUTHENTICATION False

### DIFF
--- a/masonite/auth/Auth.py
+++ b/masonite/auth/Auth.py
@@ -3,6 +3,7 @@
 import uuid
 
 import bcrypt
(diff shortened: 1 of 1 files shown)


ralliart2004:
Follow-up commit incoming. My return in login function for the token is a string when the return expects object or bool.

PR #528, opened 2019-01-19
added a new test to work with lists on request input

rfschubert:
When posting an `JSON` with an `list` it does not return the correct python `list`.

Now, it returns the `list` but as `Cellar` has one small bug, it return QS as `list` over `str`.

Before merge, need decide what to do with `parse_qs` from `Cellar` or hardcode it on `Masonite` or wait for they update it on correct way

### DIFF
--- a/masonite/helpers/misc.py
+++ b/masonite/helpers/misc.py
@@ -39,13 +39,16 @@ def dot(data, compile_to=None):
 def clean_request_input(value):
     import html
 
-    if isinstance(value, str):
-        return html.escape(value)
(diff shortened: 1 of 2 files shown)


josephmancuso:
🤔 whoops ..

josephmancuso:
not sure what i did lmao.. tried something new but i guess it didn't work

josephmancuso:
i can't reopen the PR for some reason but i just manually merged the changes to 2.1 instead. You should still get contributions for it

PR #547, opened 2019-01-24
added CORSMiddleware and test file

rfschubert:


app/http/middleware/CORSMiddleware.py
```diff
@@ -0,0 +1,34 @@
...
+    |
+    | Appends CORS headers to HTTP response
+    |
+    | Ex. cors = {"Access-Control-Allow-Origin": "*",
+    |             "Access-Control-Allow-Headers": "Content-Type",
+    |             "Access-Control-Allow-Methods": "DELETE, GET, HEAD, OPTIONS, PATCH, POST, PUT"}
+    |
+    """
+
+    CORS = {}
```
vaibhavmule: Shouldn't we have predefined headers here so that when we use this middleware we are up and running?

rfschubert: We can use predefined, but for security questions, i told it's better allow user configure only if he needs.

I do not like the way of `Django` because it allows for default `*`, so newbies normally do not configure and all server is exposed to CORS, and lots of time, this is not needed.

On my option, it's better stay `closed` and user `open` it only if has needs

rfschubert: I think also, we should be able, but don't know yet how to implement, a way to allow CORS only on specific routes, i do not know if the way `Masonite` is builded it can be done only few routes. 

But it gotta be awesome if we allow CORS only to few URL's and not all URL's.

rfschubert: if we go on `Django` style, we can configure like this

```python
CORS = {"Access-Control-Allow-Origin": "*"}
```

Personally I don't like it. But if you think it's better do it, i can commit the update. Just tell me :)

rfschubert: if we go on Django style, we can configure like this

```python
CORS = {"Access-Control-Allow-Origin": "*"}
```

Personally I don't like it. But if you think it's better do it, i can commit the update. Just tell me :)

vaibhavmule: Yes, kind of agree with you, we need to have docs in such a way. But would wait for @josephmancuso to have an opinion here. 

Regarding you want cors to be for specific routes,
You can just do.
```
    RouteGroup([
        get('/url1', ...),
        get('/url2', ...),
        get('/url3', ...),
    ], middleware=('auth', 'jwt', 'cors')),
```

rfschubert: Sure! the way you just writed is perfect.

I liked the `middleware=('auth', 'jwt', 'cors')` thing. Will be deadly simple.

rfschubert:
I'll anotate this on my to-do list and try update this soon, if i got some good changes that be near what you asked, i tell you.

i'm working on my micro service that is using `Masonite`, so maybe i can change it and do few tests

hammacktony:
We could have something in the `app/http` folder where we have a `headers.py` or something than can have all the headers we want to use. And it can have different headers for different parts of the app. The `api` can have certain ones. Front-facing requests can also have their own as well.

rfschubert:
I do not find a good approach to solve this issue... Any clues?

PR #551, opened 2019-01-27
Add migration message

### LINKED ISSUE #530: Add a check for unmigrated migrations
When you first run `craft serve -r` or `craft serve` you should be notified if you don't have a migration migrated

josephmancuso:
This PR adds the migration message when there are unmigrated migrations.

screen shot 2019-01-26 at 10 45 27 pm

Closes #530


josephmancuso:
current issue with this PR: when not connencting with the right credenntials it shows an exception instead of nothing

PR #558, opened 2019-01-30
Add new data type to be serialized after response into json.

### LINKED ISSUE #549: Add ability to return a model which returns a json response
Currently when returning a json response, we have to serialize the model first like this:

```python
def show(self):
    return User.where('key', 'value').get().serialize()
```

It seems sort of redundent to always have to serialize so we should be able to just return the model and Masonite should serialize for us:

```python
def show(self):
    return User.where('key', 'value').get()
```

This just needs a new if statement inside the `ResponseMiddleware` inside `core`

Marlysson:
@josephmancuso @vaibhavmule 
Closes #549 

Starting implementing the conditionals to verify each Response's type.

Steps that I consider to getting this done.

**IMPLEMENTATION**
- [ ] Verify each type `Model`/`Collection`

**TEST**

- [ ] Create UserMock simulating an Orator model
- [ ] Create controller test to get the method's return and verify his type.
- [ ] Verify wheter return is a json.

### DIFF
--- a/masonite/response.py
+++ b/masonite/response.py
@@ -90,9 +90,9 @@ class Response(Extendable):
         if isinstance(view, dict) or isinstance(view, list):
             return self.json(view)
         elif isinstance(view, Collection):
-            return view.serialize()
+            view = view.serialize()
         elif isinstance(view, Model):
-            return view.to_json()
+            view = self.json(view.to_dict())
         elif isinstance(view, int):
             view = str(view)
         elif isinstance(view, View):
--- a/masonite/response.py
+++ b/masonite/response.py
@@ -90,7 +90,7 @@ class Response(Extendable):
         if isinstance(view, dict) or isinstance(view, list):
             return self.json(view)
         elif isinstance(view, Collection):
-            view = view.serialize()
+            view = self.json(view.serialize())
         elif isinstance(view, Model):
             view = self.json(view.to_dict())
         elif isinstance(view, int):
--- a/masonite/response.py
+++ b/masonite/response.py
@@ -6,6 +6,8 @@ from masonite.exceptions import ResponseError
 from masonite.helpers.Extendable import Extendable
 from masonite.view import View
 
+from orator.support.collection import Collection 
+from orator import Model
 
 class Response(Extendable):
 
@@ -87,6 +89,10 @@ class Response(Extendable):
 
         if isinstance(view, dict) or isinstance(view, list):
             return self.json(view)
+        else if isinstance(view, Collection):
+            return view.serialize()
+        else if isinstance(view, Model):
+            return view.to_json()
         elif isinstance(view, int):
             view = str(view)
         elif isinstance(view, View):
--- a/masonite/response.py
+++ b/masonite/response.py
@@ -89,9 +89,9 @@ class Response(Extendable):
 
         if isinstance(view, dict) or isinstance(view, list):
             return self.json(view)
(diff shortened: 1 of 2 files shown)


masonite/response.py
```diff
@@ -87,6 +89,10 @@ def view(self, view, status=200):
 
         if isinstance(view, dict) or isinstance(view, list):
             return self.json(view)
+        else if isinstance(view, Collection):
+            return view.serialize()
+        else if isinstance(view, Model):
+            return view.to_json()
```
josephmancuso: it's `elif` not `else if` 😉

Marlysson: It's Java mind here.. So sorry.. 😃

josephmancuso: haha 😅

vaibhavmule: Line number 95. It should be `return self.json(view.to_dict())` as you want add JSON headers while returning response

josephmancuso:
idk why i thought this was in the `ResponseMiddleware` 🤔 . Must have forgotten we moved it out and put it into a response class.

josephmancuso:
as for a test, it would likely look something like:

```python
    def test_view_serialize_collection(self):

        self.response.view(MockUser().all())
        assert self.app.make('Response') == "{'1', '2'}"
```

or whatever that json response is of `Collection([1,2])` is

Marlysson:
Don't necessary instantiate a Controller and get the `Response` returned by some method ? 

Lol, easier than I thought..

masonite/response.py
vaibhavmule: You would want to return json with json headers. this line should be. 

```suggestion
            return self.json(view.to_dict())
```

josephmancuso: oh good catch

Marlysson: The View has the `to_dict` method?

josephmancuso: yes it should be to_dict() since the json method needs a dictionary

masonite/response.py
```diff
@@ -87,6 +89,10 @@ def view(self, view, status=200):
 
         if isinstance(view, dict) or isinstance(view, list):
             return self.json(view)
+        elif isinstance(view, Collection):
+            view = view.serialize()
```
josephmancuso: looking at this again, this line should also be JSON `view = self.json(view.serialize())`

tests/test_response.py
```diff
@@ -54,3 +67,13 @@ def test_view_can_set_own_status_code(self):
...
+
+    def test_view_should_return_a_json_response_when_retrieve_a_user_from_model(self):
+        
+        assert isinstance(MockUser(), Model)
+
+        self.response.view(MockUser().find(1))
+
+        json_response = [{'name':'TestUser', 'email': 'user@email.com' }]
+
+        assert self.app.make('Response') == json_response
```
josephmancuso: this test is good but not correct. 

```python
json_response = "[{'name':'TestUser', 'email': 'user@email.com' }]"
assert self.app.make('Response') == json_response
```

that should be the correct assertion.

josephmancuso: since it needs to be a string when going to the browser

josephmancuso: That change in the response class above should fix that.

josephmancuso:
pulled this in, made a few minor tweaks and merged into master. 🎊 🎉 . Thanks for contributing!

89aa00f5e8c236e89c6578a25ca310ad026f5586

PR #559, opened 2019-01-30
Add routes and controller to route initializers

### LINKED ISSUE #555: Change how route classes are constructed
Currently we have something like this:

```python
Get().route('/route', 'Controller@show').middleware(..)..
```

but it really should be:

```python
Get('/route', 'Controller@show').middleware(..)..
```

This should be backwards compatible. For example the constructor of `Get()` should be:

```python
def __init__(self, route, controller):
    self.route(route, controller)
```

but it just seems odd to declare `Get().route(..)` and not `Get(..)`

josephmancuso:
This PR adds the ability to specify routes and controllers in the constructor instead of the normal `.route()` method like so:

```python
ROUTES = [
    Get().route('/some/url', 'Controller@show').middleware(..)
]
```

now becomes:

```python
ROUTES = [
    Get('/some/url', 'Controller@show').middleware(..)
]
```

Both ways are still possible.
Closes #555

### DIFF
--- a/masonite/routes.py
+++ b/masonite/routes.py
@@ -360,25 +360,29 @@ class BaseHttpRoute:
 class Get(BaseHttpRoute):
     """Class for specifying GET requests."""
 
-    def __init__(self):
+    def __init__(self, route=None, output=None):
         """Get constructor."""
         self.method_type = ['GET']
         self.list_middleware = []
+        if route and output:
+            self.route(route, output)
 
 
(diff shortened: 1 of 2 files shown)


josephmancuso:
we could. Wouldn't make a difference though. This addition actually uses the `.route()` method in the initializer still. I'm not 100% sure if i wanna remove the `route` method yet in 2.2

PR #563, opened 2019-02-05
reworked queue connection drivers

josephmancuso:
This PR starts to rework a bit of the queue features to be able to support more features and drivers in the future


josephmancuso:
Note to myself or anyone following this PR: I am attempting here to make the queue worker more modular and able to call any driver here. The only issue i am seeing (and i need further testing) is that if we put the connection details in the queue worker, it needs to connect every single time because the driver is instantiated everytime a set of jobs are pushed to the queue (every time to the `.push()` method is called). I am wondering if there is a way to design this that we keep the connection open ... 

I am not sure that these changes introduced this issue or if this was still an issue before (rapid connecting to the message broker every push). 

I'll have to test more on this.

josephmancuso:
This PR introduces a whole bunch of new features:

* jobs that fail will run again 3 times before being handled
* jobs can now have a `failed` method on them to use as a callback for when a job fails. Like sending an email.

## New Commands

You may now use 2 new commands: `queue:failed` and `queue:table`. The `queue:table` command will create a new migration based on a failed jobs table and the `queue:failed` command will run all the commands 

Also now has a bit better UI to seeing job statuses as they come in
screen shot 2019-02-06 at 10 54 57 pm

PR #576, opened 2019-02-12
Masonite 2.2 - Next Major (Beta 3)

josephmancuso:
- [x] Means whether or not the feature has been documented.

## Running Changelog

- [x] Added a `/` to the front of a route URL if it does not exist.
- [x] Added URL parameters are now added to the controller method.
- [ ] Added storage drivers to manage files
- [x] Changed the async driver to use either threading or multiprocessing
- [x] Added new HTTP verbs to routes (OPTIONS, TRACE, CONNECT etc)
- [x] Errors and 404's are now returned in JSON if the incoming content type is `application/json`
- [x] Added parameters to middleware in routes
* Rearranged drivers all into own packages

☝️ Beta 1 release
****

- [x] Craft serve command now reloads by default. You can still pass in the `-r` flag but it is not required. You can also disable reloading by using `craft serve -d` for `--dont-reload`
- [x] Added a `accept(*)` to the upload drivers 
- [x] Added a lot more view helpers 
- [x] Changed all tests to unittests
- [x] Added better test case for database tests

☝️ Beta 2 release

****
- [x] Added validation library
- [x] view back helper now defaults to the current path
- [x] Auth class now is resolved and does not need a request to be passed in

☝️ Beta 3 Release
****

- [x] Added register method to auth class
- [x] Changed regex compiling to before the server starts
- [x] Added container remembering
- [x] Added redirection `with_errors` 
- [x] Added publishing 
- [x] Added JWT auth driver 

☝️ Beta 3 POST releases

- [x] Added test scaffold command, publish command
- [x] Removed all helper functions for routes
- [x] Moved all configs into lazy loading
- [x] Added show method to contains
- [x] Added singleton to the container class 
- [x] Added `json` test 
- [x] Added environment for testing
- [x] Added with and without CSRF 

☝️ Extras


vaibhavmule:
just curious how much is validation library loc would be?

josephmancuso:
not much. Prob around 800 LOC

vaibhavmule:
So we are getting rid of pytest and validation.py in this PR....

josephmancuso:
@vaibhavmule well yes and no. We are getting rid of pytest tests but we can still run tests with pytest. We are just changing how tests are created. I like the approach of teardown and setup methods rather than fixtures a lot better than

.travis.yml
```diff
@@ -10,15 +10,15 @@ matrix:
...
+      script: travis_retry coverage run -m pytest && coveralls && coverage xml && curl https://deepsource.io/cli | sh && ./bin/deepsource report --analyzer test-coverage --key python --value-file ./coverage.xml
 
 install:
 - pip install masonite_cli
 - pip install -e .
 - pip install -r requirements.txt
 
 script: 
-- travis_retry coverage run -m pytest
+- travis_retry coverage run -m pytest && coveralls
```
vaibhavmule: Are we running coveralls for every version of the python?

josephmancuso:
Yeah I just lazy loaded all configs so we can use and import Masonite into other packages and not break because of lack of config files

tests/testing/test_route_tests.py
```diff
@@ -51,3 +52,11 @@ def test_can_have_user(self):
 
     def test_json(self):
         self.assertTrue(self.json('POST', '/unit/test/json', {'test': 'testing'}).contains('testing'))
+
+    def test_patch(self):
+        self.assertTrue(self.patch('/unit/test/patch', {'test': 'testing'}).contains('testing'))
+
+    def test_csrf(self):
+        self.with_csrf()
+        with self.assertRaises(InvalidCSRFToken) as context:
```
vaibhavmule: `context` is unused here, on purpose?

josephmancuso:
@vaibhavmule because for some reason unittest uses camel case and if we use PEP 8 and use underscore its gonna confuse developers on which methods are underscore and which are camelcase. So I think its better for consistency to just use camelCase to keep the flow of unittest

vaibhavmule:
@josephmancuso oh I didn't thought through this. But this is good that you noticed

PR #582, opened 2019-02-14
add -m to craft model

### LINKED ISSUE #569: Create migration file with craft model command
```
$> craft model Photo
$> craft migration create_photos_table
```
Just thinking about a little way to speed up generating of database stuff like Artisan. So we can add a -m parameter.
```
$> craft model Photo -m
```
And this will create a migration file automatically 👍

vaibhavmule:
Closes #569

### DIFF
--- a/masonite/commands/ModelCommand.py
+++ b/masonite/commands/ModelCommand.py
@@ -1,15 +1,45 @@
 """New Model Command."""
-from masonite.commands import BaseScaffoldCommand
+import subprocess
+
+from masonite.app import App
+from masonite.helpers.filesystem import make_directory
+from masonite.view import View
+
+from cleo import Command
 
 
-class ModelCommand(BaseScaffoldCommand):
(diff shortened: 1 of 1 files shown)


masonite/commands/ModelCommand.py
```diff
@@ -1,15 +1,45 @@
...
+            f.write(
+                view.render(self.template, {
+                            'class': class_name.split('/')[-1]}).rendered_template
+            )
+            self.info('{} Created Successfully!'.format(self.scaffold_name))
+            f.close()
+
+        if self.option('migration'):
+            model_name = class_name.lower() + 's'
+            subprocess.call(['orator', 'make:migration', 'create_{}_table'.format(model_name),
```
vaibhavmule: I tried doing `self.call` but it was giving a weird error, `[TypeError]
  can only concatenate list (not "str") to list` that I wasn't able to debug, that is why I used subprocess.

Issue #567, opened 2019-02-10
Multiprocessing in queue async driver.


hammacktony:
I think it would be good to have the ability to specify whether the user wants to use threading or multiprocessing as their async method. Have a parameter that the user can specify.

josephmancuso:
is there a benefit in multi processing vs multi threading?

hammacktony:
They have different use cases. 

Essentially:
- Threads are spawned from one process. 
- Multi-process...well, use more than one.

IMO, it really depends what the dev is trying to do.

josephmancuso:
This should be in the next major.

This might be a good built in package to use https://docs.python.org/3/library/concurrent.futures.html

josephmancuso:
maybe have a config like this:

```python
DRIVERS = {
    'async': {
        'type': 'process' # or 'thread'
    }
}
```

PR #589, opened 2019-02-23
Update and expand async queue driver

hammacktony:
Have option to use threading or multiprocessing. Also uses `concurrent.futures` for a consistent api between threading and multiprocessing. Also allows the user to specify the amount of threads/processes they want to use.

### DIFF
--- a/masonite/drivers/QueueAsyncDriver.py
+++ b/masonite/drivers/QueueAsyncDriver.py
@@ -10,6 +10,7 @@ from masonite.exceptions import QueueException
 from concurrent.futures import ThreadPoolExecutor
 from concurrent.futures import ProcessPoolExecutor
 
+
 class QueueAsyncDriver(BaseQueueDriver, QueueContract):
     """Queue Aysnc Driver."""
 
@@ -21,11 +22,10 @@ class QueueAsyncDriver(BaseQueueDriver, QueueContract):
         """
         self.container = app
 
-
     def _threading(self):
         """ Implements Async by Threading """
 
-        with ThreadPoolExecutor(max_workers = self.workers) as executor:
+        with ThreadPoolExecutor(max_workers=self.workers) as executor:
             for obj in self.objects:
                 if inspect.isclass(obj):
                     obj = self.container.resolve(obj)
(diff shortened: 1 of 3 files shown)


hammacktony:
There is a bug in Python 3.4 `concurrent.futures`.

hammacktony:
`concurrent.futures.ThreadPoolExecutor` does not play nicely with Python 3.4. Since they are not accepting any changes except security patches, I will close this PR. This will be revisited when 3.8 comes out and we do not support 3.4 any longer.

josephmancuso:
wait so is this good now?

masonite/drivers/QueueAsyncDriver.py
```diff
@@ -19,22 +23,60 @@ def __init__(self, app: App):
...
+                if inspect.isclass(obj):
+                    obj = self.container.resolve(obj)
+
+                try:
+                    executor.submit(
+                        fn=getattr(obj, self.callback), args=self.args, kwargs=self.kwargs)
+                except AttributeError:
+                    # Could be wanting to call only a method asyncronously
+                    executor.submit(fn=obj, args=self.args, kwargs=self.kwargs)
+
```
josephmancuso: We can break the 2 methods above into a single method because the only difference is that one line. So we can make that a variable of `self.processor` for example and we can set that in the push method or more preferably make it a special `self._get_processor()` method since we may need to set some special cases like for Python 3.4 and threading

masonite/drivers/QueueAsyncDriver.py
```diff
@@ -19,22 +23,60 @@ def __init__(self, app: App):
...
+    def push(self, *objects, args=(), kwargs={}, callback='handle', mode='threading', workers=None):
```
josephmancuso: This has been in the back of my head for a while .. but i think we should experiment with making these `**kwargs` here. we may need to change `kwargs` here to `keywords` and then make a `**kwargs` at the end. Does that make sense? Or something similiar.

Here is an example definition:

```
    def push(self, *objects, args=(), kwargs={}, **keyword_args):
        """Push objects onto the async stack.

        Arguments:
            objects {*args of objects} - This can be several objects as parameters into this method.
        """

        self.workers = keyword_args.get('workers', None)
        self.objects = objects
        self.callback = keyword_args.get('callback', 'handle') # get callback keyword or `handle` as default argument
        self.args = args
        self.kwargs = kwargs
```

josephmancuso: reason being that a driver may need something weird like `hop=3` or something wacky. If we have `**keyword_args` then all drivers will simply ignore that but the actual driver that needs it will be able to use it.

hammacktony:
Closes #567

josephmancuso:
this looks really good. I'd merge it but i want to manually test it when i get some more free time

PR #593, opened 2019-02-24
Add view shortcut

### LINKED ISSUE #592: Easier way to set status code
Right now, we set the status code by `request.status_code(###)` in the controller. This does feel clunky. The more I think about it, the more I really think we should implement something like 
```return view('page.html'), 201```. 

Doing,

```
`request.status_code(###)` 

return view('blah')
```

is very cumbersome.

We can also set it in the `view` method. Either way, I think this change could really help devs out. It is just simple.

### PR DESCRIPTION
**Pull request by josephmancuso**
Closes #592 

This PR allows you to set a status codes easier. You can now do something like this when returning a view.

we currently do something like this:

```python
def show(self, view: View, request: Request):
    request.status(203)
    return view.render('some.view', {'key': 'value'})
```

now you can do:

```python
def show(self, view: View):
    return view.render('some.view', {'key': 'value'}) & 203
```

the `&` is a shortcut is still experimental and could change. Any suggestions are welcome.

### DIFF
--- a/app/http/controllers/TestController.py
+++ b/app/http/controllers/TestController.py
@@ -4,6 +4,7 @@ from masonite.exceptions import DebugException
 from masonite.request import Request
 from masonite import Queue
 from app.jobs.TestJob import TestJob
+from masonite.view import View
 
 
 class TestController:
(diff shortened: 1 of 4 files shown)


hammacktony:
Can we also set the status? Like `return view(##) & '404 Not Found'`. There are probably some that would like that.

Marlysson:
Why not a third argument like:

view.render(template, data, status)

josephmancuso:
@Marlysson thats a fair argument. we could deff do that. Would prob make more sense honestly.

hammacktony:
What if you were returning something else than a view?

hammacktony:
I also think that `View` objects should not mess with `Request` objects. There should be some separation. These two objects meet in the controller. So, this is where this should be implemented.

josephmancuso:
the more i think about this, the more i'm not really a fan .. Not sure if this is the way to accomplish what we are trying to do ..

Issue #597, opened 2019-02-25
Socket IO websockets


hammacktony:
I know a lot of people use Socket IO for their websockets. I think it wise to create a driver so that others can implement it in there applications.

hammacktony:
I do not think this is pressing issue. We will revisit this at another time.

PR #598, opened 2019-02-26
Socketio websockets

hammacktony:
This PR is in response to issue #597. Goal is to have the ability to use socketio in masonite applications.


josephmancuso:
does this need to have a local server running? We may need to make a command for that. Similiarly to how we restructured the queue command.

hammacktony:
Yeah. Most of the socket.io websocket servers are written using NodeJS. There are some Python ones, but they are pretty barebones. I am debating on writing one because you to manually configure it. There is some server configuration that cannot be done via a cli or environment variables

hammacktony:
This is not a high priority and is far away from being merged. Will revisit at a later date.

Issue #602, opened 2019-03-02
Add a compact function


josephmancuso:
This function will take the varialble in the current scope and return it as a dictionary along with the value as the key. For example it will turn this:

```python
        sites = Site.where('user_id', 1).get()
        current = Site.where('user_id', 1).first()
        deployments = Deployment.where('site_id', current.id).get()
        server = Server.where('site_id', current.id).get()

        return view.render('index', {'sites': sites, 'deployments': deployments, 'current': current, 'server': server})
```

into this:

```python
        sites = Site.where('user_id', 1).get()
        current = Site.where('user_id', 1).first()
        deployments = Deployment.where('site_id', current.id).get()
        server = Server.where('site_id', current.id).get()
        return view.render('index', compact('sites', 'current', 'deployments', 'server'))
```

PR #603, opened 2019-03-02
added compact helper

josephmancuso:
it is very repetitive when you want to have a variable name and you want to send a variable to the view with the same name. This may look something like this:

```python
def show(self, view: View):
    posts = Post.all()
    users = User.all()
    return view.render('some.template', {'posts': posts, 'users': users})
```

now you can do:

```python
from masonite.helpers import compact

def show(self, view: View):
    posts = Post.all()
    users = User.all()
    return view.render('some.template', compact(posts, users))
```

You can also pass in a dictionary which will update accordingly:

```python
from masonite.helpers import compact

def show(self, view: View):
    posts = Post.all()
    users = User.all()
    user_blogs = Blog.where('user_id', 1).get()
    return view.render('some.template', compact(posts, users, {'blogs': user_blogs}))
```

### DIFF
--- a/masonite/exceptions.py
+++ b/masonite/exceptions.py
@@ -77,5 +77,6 @@ class DumpException(Exception):
 class ViewException(Exception):
     pass
 
+
 class AmbiguousError(Exception):
     pass
--- a/masonite/exceptions.py
+++ b/masonite/exceptions.py
@@ -76,3 +76,6 @@ class DumpException(Exception):
 
 class ViewException(Exception):
     pass
+
+class AmbiguousError(Exception):
+    pass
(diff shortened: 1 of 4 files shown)


vaibhavmule:
Do you mean?
posts and users without stirng?
`return view.render('some.template', compact(posts, users))`

josephmancuso:
#602

nioperas06:
Is this available in `2.1.19`?

josephmancuso:
no it will be in 2.1.20. I can release now if you like :)

nioperas06:
🙌

josephmancuso:
released

PR #611, opened 2019-03-06
S3 Storage Driver

arpitgogia:
A very simple and basic implementation of an S3 Storage Driver.
Here's the [API reference](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#id200)

@josephmancuso I have only implemented methods that are pretty straight forward to wrap around for now. Methods like `append` and `move` do not seem to have direct APIs. Happy to discuss the implementation strategy for the same. 

Certain methods (e.g. `put` and `upload`) seem redundant in this case, can be handled by raising a `NotImplementedException` and directing to the corresponding implemented method.


vaibhavmule:
Can you tell me what are you trying to implement here?

josephmancuso:
He’s implementing an S3 version of the disk storage driver. I’d like to create these wrappers around managing files and created one for disks he is going to create one for the S3 driver

josephmancuso:
not sure whats going on with this but i don't see any files changes 🤔 . I also can't find the class you made. It looks like the last commit deleted the class

arpitgogia:
@josephmancuso sorry, I missed your comment above about the file getting deleted. What’s the solution for this? I still have the file locally because I haven’t pulled.

josephmancuso:
not sure how the file was deleted but it's possible that a simple `git reset <commit>` . can get you back to where you were before the file was deleted.

arpitgogia:
@josephmancuso I think it's there now

PR #615, opened 2019-03-07
patch for issue 612

nilsreichert:
added assert to two lines under function test_upload_manager_changes_accepted_files in test_upload_manager.py


josephmancuso:
thanks for contributing ! ❤️

Issue #622, opened 2019-03-10
Allow user to define output of 404 and 500 status codes


hammacktony:
This issue is addresses how when you set either a 404 or 500 status code, you get the automatic html response. This is fine for frontend dev, but not good for api development. I want to set either a 404 and 500, and give a custom response. 

I think I have asked this before in Slack. I cannot remember if we have already addressed this.

vaibhavmule:
+1 for this as I feel this should be configurable!

josephmancuso:
wait .. i just realized .. this is possible for views. But not yet for json so i'll keep this open.

I coded this when i coded the exception stuff. 

https://github.com/MasoniteFramework/core/blob/2.1/masonite/providers/StatusCodeProvider.py#L10-L23

josephmancuso:
notice how it first checks for a `errors/xxx` template in your project and renders that if it exists. Else it will render the Masonite provided one.

hammacktony:
Yeah. I was referring to json

josephmancuso:
@MasoniteFramework/maintainers anybody have any ideas on how we should go about this? we basically just need to change what is displayed in the exception handler but how should this be configurable?

josephmancuso:
if we made this a middleware ... We could then do something like this in the routes:

```python
ROUTES = [
    Get(..).middleware('exception:json')
]
```

josephmancuso:
talking to @hammacktony we could just look at the headers ... duh .. if the headers is `application/json` then output a json response

vaibhavmule:
I do not like this approach of checking data type, but also I do not have a solution for this at the moment!

PR #631, opened 2019-03-17
Feature/json 404 and 500 response

hammacktony:
Closes Issue #622.

Allows one to specify either to return a json response or html response for 404, 405, and 500 status codes.


hammacktony:
Right now, it checks for a content type of `application/json`.  We might want expand this to include other types, like xml.

Please feel free to comment.

josephmancuso:
Made some updates. Here is a sample of what the response will look like

Screen Shot 2019-03-17 at 4 10 53 PM
Screen Shot 2019-03-17 at 4 10 41 PM

PR #663, opened 2019-04-01
collect alias ADDED

danieldsf:
I have added the Collection class as collect to the helpers import management file. (__init__)


danieldsf:
I just added a test as suggested before.

PR #658, opened 2019-03-31
Sign related view helpers ADDED

danieldsf:
I just created a new file that exposes the sing related functions like sign, unsign, encrypt and decrypt.

PR #665, opened 2019-04-01
View Helpers ADDED

danieldsf:
As said before in the PR #658. This PR aims to provide shortcuts for some methods that can be used for views. I think I added all the requested ones.


vaibhavmule:
Nice.  These are really good helpers

PR #679, opened 2019-04-15
Add better container making

### LINKED ISSUE #678: Change how making an object from the container resolves
The entire point of a service container is to manage application dependencies automatically.

If i have an object like this:

```python
class Mail:

    def __init__(self, request: Request, mail: Mail):
        self.request = request
```

i should be able to load this class into the container and have it resolve dependencies when I make the class from the container at that time.

For example, I should be able to do this:

```python
app.bind('Mail', Mail)

mail = app.make('Mail').request == Request
```

So the class is actually initialized WHEN the class is retrieved from the container.

josephmancuso:
This PR is a breaking change of how `making` from the container works.

before, if you bound a class into the container, you got the class back. But now if you `make` a class from the container, you get back an object.

Basically anytime you `make` and the object is a class it will now do this:

```
obj = app.resolve(app.make('Class'))
```

It will do this under the hood now.

### Singletons

The container now has a `singleton` method:

```python
app.singleton('Request', Request)
```

which is just a wrapper around:

```python
app.bind('Class', app.resolve(Obj))
```

It will initialize the class into an object and bind the object into the container. All calls made to the class will return the same object.

Closes #678 #680


josephmancuso:
There should be a way to also "make" the class without initializing it. For example if you need to modify the class on the fly without turning it into an object before you get it.

Maybe we will call this `grab()` instead of `make()`

josephmancuso:
Changed how each parameter resolves. Now Masonite will look at each parameter before it resolves and decides on if it should resolve it.

josephmancuso:
ran benchmarks on this and did not see any notable differences on speed. In fact I saw a bunch of benchmarks scores at an all time high but those can just be the statistical outliers

PR #683, opened 2019-04-22
added new url params to route

### LINKED ISSUE #682: Add changes to request.route() method
Right now if we go to a route like:

```python
Get('/dashboard/@site/deployments', 'SomeController@show').name('dashboard.show')
```

and then we need to add a named route in our templates:

```html
<a href="{{ route('dashboard.show', {'site': 1}) }}">
```

This seems a little repetitive .. We already have that dictionary inside the URL params like `request.param('site')`.

Since we already have that information, we SHOULD be able to exclude the extra parameters and it should update automatically for us like this:

```html
<a href="{{ route('dashboard.show') }}">
```

josephmancuso:
Closes #682

### DIFF
--- a/masonite/request.py
+++ b/masonite/request.py
@@ -714,7 +714,11 @@ class Request(Extendable):
         if full:
             route = application.URL + self._get_named_route(name, params)
         else:
-            route = self._get_named_route(name, params)
+            try:
+                route = self._get_named_route(name, params)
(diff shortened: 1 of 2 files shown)


vaibhavmule:
What is this about?

josephmancuso:
@vaibhavmule #682

PR #690, opened 2019-05-03
added tests for new validation library

### LINKED ISSUE #689: validation
Need to write my own validation library .. Too many features to add that would be awesome, the one we have is just not enough and the wrapper we have around it needs to be updated to the point where it's not salvageable.

josephmancuso:
Closes #689


josephmancuso:
PR is a WIP. A couple of things we still need to figure out is:

1. How will we use this inside the controller?

Thinking of adding these rules right to the request class:

```python
def show(self, request: Request):
    validated = request.validate({
        'user': 'required|length:3,8',
        'password': 'required|length:4,8|strong'
    })

    if validated.errors():
        return request.back(with_errors=True)
        # or
        return request.redirect(name='error.page', with_errors=True)
        # or
        return request.back().with_errors()
```

The other thing is adding a validation class to the container / the flow of Masonite. If i want to use this inside my controller:

If i want to add a custom validator class like `haveibeenpwned`

```python
def show(self, request: Request):
    validated = request.validate({
        'password': 'required|haveibeenpwned'
    })
```

then we need to add the logic to the container:

```python
from masonite.validation import Validator
from app.validators import HaveIBeenPwned

def boot(self, validator: Validator):
    validator.add(HaveIBeenPwned)
```

Something like that? So then when it checks for validation it will also check for this new one. This way packages can add their own validation and the validation feature will be super extendible

josephmancuso:
This actually also needs to return several errors messages, not just one

josephmancuso:
I really like the style of `length:3..8` over `length:3,8`

josephmancuso:
Also need to be able to validate nested parameters like this:

```php
$request->validate([
    'title' => 'required|unique:posts|max:255',
    'author.name' => 'required',
    'author.description' => 'required',
]);
```

For validating incoming JSON for example

josephmancuso:
Available validators:

- [ ] `required`
- [ ] `numeric`
- [ ] `string`
- [ ] `none`
- [ ] `length`
- [ ] `in_range`
- [ ] `equals`
- [ ] `contains`
- [ ] `is_in`
- [ ] `greater_than`
- [ ] `less_than`
- [ ] `isnt`
- [ ] `truthy`
- [ ] `json`

josephmancuso:
need an opposite of when

```python
.when_not(
    equals('isManager', 1)
).then(
   greater_than('age', 18),
   accepted('terms')
)
```

josephmancuso:
added date validations:

* is_past
* is_future
* before_today
* after_today

josephmancuso:
need to add:

* `ip` for ip address

PR #702, opened 2019-05-16
max line length to 99

vaibhavmule:


vaibhavmule:
That was the fastest PR merge I have ever seen...

Issue #694, opened 2019-05-07
Change auth class


josephmancuso:
There is a new way that classes are being resolved in 2.2 (currently in the develop branch).

now instead of doing this:

```python
from masonite.auth import Auth

def show(self, request: Request):
    Auth(request).login(..)
```

we can now do this:

```python
from masonite.auth import Auth

def show(self, auth: Auth):
    auth.login(..)
```

This is because now parameters that are classes (`<class masonite.auth.Auth>` type objects) are now resolved out of the container **as they are resolved**. Meaning we can resolve the auth constructor as it is resolved from the parameter list. This allows us to automatically resolve the auth initializer which will pass in the request class before we get the object for us.

PR #709, opened 2019-05-17
create auth.register method

vaibhavmule:


vaibhavmule:
Note: This needs to be to do after https://github.com/MasoniteFramework/core/issues/694

josephmancuso:
With the new `DatabaseTestCase` class we can add legit database tests. I added one and found an issue with one of your lines. you were doing a `==` instead of an assignment (`=`)

josephmancuso:
not a problem though :) thats what tests are for

PR #713, opened 2019-05-19
Feature/534

josephmancuso:
This PR introduces the concept of Provider Publishing. This is the concept of a provider being responsible for adding files to an application like migrations or routes or templates.

### DIFF
--- a/masonite/packages.py
+++ b/masonite/packages.py
@@ -20,14 +20,9 @@ def create_or_append_config(location, name=False):
         print('\033[92mConfiguration File Created!\033[0m')
     else:
         # Append to the file
-        # tempfiles is a list of file handles to your temp files.
-        project_config = open(config_directory + '/' + file_name, "a")
-        package_config = open(location, 'r')
+        with open(config_directory + '/' + file_name, "a") as project_config, open(location, 'r') as package_config:
+            project_config.write(package_config.read())
 
-        project_config.write(package_config.read())
-
-        project_config.close()
-        package_config.close()
         print('\033[92mConfiguration File Appended!\033[0m')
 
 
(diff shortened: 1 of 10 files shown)


josephmancuso:
You can use this by using a new publishing method on a provider:

```python
class SomeProvider:

    def boot(self):
        self.publishes({
            '/from/file': '/to/file'
        })
```

you can also tag this so you may want to group together multiple files under a namespace like `views`:

```python
class SomeProvider:

    def boot(self):
        self.publishes({

            '/package/views/admin.html': 'resources/templates/vendor/admin.html',
            '/package/views/dashboard.html': 'resources/templates/vendor/dashboard.html',
            '/package/views/users.html': 'resources/templates/vendor/users.html',
            '/package/views/pages.html': 'resources/templates/vendor/pages.html',

        }, tag='views')
```

You can then use this by running a new `publish` command:

```
$ craft publish SomeProvider --tag views
```

josephmancuso:
since migrations need special logic (create the file with the correct timestamp of right now) you need to run:

```
class SomeProvider:

    def boot(self):
        self.publishes_migrations([
            '/migration-directory/1',
            '/migration-directory/2'
        ], tag='views')
```

masonite/packages.py
```diff
@@ -47,6 +47,17 @@ def append_web_routes(location):
     print('\033[92mroutes/web.py File Appended!\033[0m')
 
 
+def append_file(from_location, to_location):
+    from_fp = open(from_location, "r")
+    to_fp = open(os.path.join(os.getcwd(), to_location), 'a')
+
+    to_fp.write(from_fp.read())
```
vaibhavmule: Why can't we just move file here?

josephmancuso: sometimes we will need to append it. Think of the routes.py file. Can't just move the file, the contents of it needs to be appended at the end

vaibhavmule: Got it, thanks.

Marlysson: Which means "fp" at the end of variables' name?

josephmancuso: file pointer.

I hate making short variable names lol. I should make them longer and spell out file_pointer

masonite/packages.py
```diff
@@ -46,6 +46,16 @@ def append_web_routes(location):
     package_routes.close()
     print('\033[92mroutes/web.py File Appended!\033[0m')
 
+def append_file(from_location, to_location):
```
hammacktony: I believe you should use a context manager. Not only is it the standard way to mess with files, I believe it is also cleaner.

```python
with open(from_location, "r") as from_fp, open(to_location, 'a') as to_fp:
    to_fp.write(from_fp.read())
```

josephmancuso: actually didn't know you can use context managers like that

PR #718, opened 2019-05-22
add masonite.db to .gitignore

vaibhavmule:


clsource:
hello. I just need a little context.
Why this file need to be ignored by git?

Thanks :)

vaibhavmule:
Because when I am using it my database records might be different from other person using it..  Also it has been added for running tests and all, and versioning for that.masonite.db file is not required.  And,  I should never upload database data in versioning. 
Correct me if I am wrong.

clsource:
I'm just wondering if this file is autogenerated or is something that the user can modify.
If is autogenerating then no problem to ignore this file, if the user might want to save the records somehow
then adding a comment before the ignore as a recommendation could be useful :)

vaibhavmule:
It is file generated by sqlite.

clsource:
seems fine to me 👍

Issue #728, opened 2019-05-27
General Optimizations


hammacktony:
In response to Masonite's rankings in [https://github.com/vaibhavmule/web-frameworks](https://github.com/vaibhavmule/web-frameworks), we should aim to optimize the current code base.

Masonite is in a position where it is stable, and the only things we are working on are new features. I believe it is a good time to take stock of the current code, and optimize where we can. 

There is a lot of low hanging fruit out there. For instance,
```
# startswidth method
212 ns ± 5.85 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
# grab from beginning of list
57.3 ns ± 0.477 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
# endswidth method
216 ns ± 7.67 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
# grab from end of list
64.6 ns ± 1.79 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

Simple optimizations will add up and make the project more performant. 

Plan of attack:

- [x] - String manipulations optimizations
- [x] - Move regex compiling to boot time instead of runtime
- [ ] - Look into `__slots__`
- [ ] - Optimize Route Provider (Addressed in #727. Targeted for version 2.3)

If there are any other potential optimizations to be made, please detail so that we can work on them.

PR #731, opened 2019-05-27
Feature/728 general optimizations

hammacktony:
After some analysis, accessing the beginning and the end of a list is faster than using the `startswith` and `endswith` methods, given that the string we want to find has length one.

Reference: #728


hammacktony:
There is a possibility of having an `IndexError` if there are empty strings, but from what I can tell, that should not happen.

masonite/app.py
```diff
@@ -207,15 +207,15 @@ def collect(self, search):
...
                             providers.update({key: value})
-                    elif search.endswith('*'):
-                        if key.startswith(search.split('*')[0]):
+                    elif search[-1] == '*':
+                        if key[0] == search.split('*')[0]:
                             providers.update({key: value})
                     elif '*' in search:
                         split_search = search.split('*')
-                        if key.startswith(split_search[0]) and key.endswith(split_search[1]):
+                        if key[0] == split_search[0] and key[-1] == split_search[1]:
```
josephmancuso: This part isn't necessarily going to work. Since the search string is going to be something like `Sentry*Hook` you can't just splice the first and last letters

masonite/routes.py
```diff
@@ -149,10 +149,19 @@ def route(self, route, output):
...
+
+        Raises:
+            RouteException -- Thrown when the incoming route is an empty string.
         """
+        # Checked to make sure the route is not empty
+        try:
+            if route[0] != '/':
+                route = '/' + route
+        except IndexError:
+            raise RouteException("Undefined Route")
```
josephmancuso: whats all this doing? All that to make sure the route is not empty? can't you just do `if not route` to check for empty?

hammacktony: True.

hammacktony: All I added was the try/catch. The rest of it is what was there.

josephmancuso: oh i see now

hammacktony: I am not too happy with the fix, so this is all prone to change.

hammacktony: I think catching that edge case is important. I can see causing issues for the `"/"` routes. We all have fat fingers and can easily code `""` instead.

josephmancuso: yes plus the bottom one is valid in 2.2. Hence why you see `route = '/' + route` to begin with since its putting one there for you when you forget

josephmancuso:
is there more to this PR or are the above fixes sort of a first step let's merge this first thing?

hammacktony:
I was planning for this to be longterm. It is going to take me a bit to walk through the code base and profile things.

josephmancuso:
Ok so you're looking at 2.3 then?

hammacktony:
Yeah.

PR #746, opened 2019-06-08
Deepsource and docstrings fixes

vaibhavmule:


vaibhavmule:
@josephmancuso do you think we should `deepsource` only `masonite` folder? just like we do it for `flake8`?

vaibhavmule:
@sanketsaurav is there any way if we can run deepsource only on some folder?

Issue #617, opened 2019-03-08
Craft Dev Commands


hammacktony:
I propose a new craft command called `test`. Running tests with `pytest` is simple enough. But, I see many PRs fail due to `flake8` errors. So, why not create a convenience command that runs both. It will simulate what Travis does.

vaibhavmule:
This is a good idea, even before doing PR we can just run the command and raise PR error-free. 

I'm not sure, but I feel this should not in `craft` as it is `core` repo specific use case.

hammacktony:
I wasn't too sure where to post this issue because it could have been in either place.

I know `tox` could do this, but I feel that is overkill.

vaibhavmule:
I haven't work with `tox` much. We can use `makefile` too, and use those command in Travis too.

hammacktony:
great idea

kentaro0919:
I think "craft" is used in 2 ways. One from masonite_cli and the other from the project. I like to have project craft command to generate test.py files. And a wrapper command ``` $ python -m pytest``` for masonite_cli.

hammacktony:
Its not just the wrapper for the test commands. I forget to run `flake8`. And I have my PRs fail over things I should have fixed. When I look at a lot of others PRs that fail, they have those issues. If we can replicate (as much as we can), then our local dev experience will be smoother.

josephmancuso:
completely agree on all this

vaibhavmule:
Note: This is done but @kentaro0919 commented about a wrapper command over `python -m pytest` is still pending.

PR #748, opened 2019-06-08
add Makefile

### LINKED ISSUE #730: Makefile
I think having a Makefile would be beneficial. It can help devs lint and test (because we all know remembering those commands are hard). Also, `make` is on almost every operating system. It is a standardized way of manipulating projects.

It should,

- Clean Project (remove `__pycache__/` and other useless decorrectories) `make clean`
- Lint with Flake8 `make lint`
- Test with Pytest `make test`
- Install application. Instead of downloading `masonite-cli` and then running `craft install`, have `make install` cover any initial project dependencies

If there are any other commands which might be useful, comment below.

vaibhavmule:
Fixes #730 #617


josephmancuso:
I don’t think the Travis script is running the requirements anymore

vaibhavmule:
@josephmancuso I thought same, but it was working, Please have a look at PR and sorry for dozens of commit, you might want to squash and merge this.

PR #751, opened 2019-06-08
Fix masonite version and refactor setup.py

vaibhavmule:
- Masonite version
```
import masonite
masonite.__version__ # this follows https://www.python.org/dev/peps/pep-0396/
```

 - fixed project description, which looks empty at https://pypi.org/project/masonite/
- add python_requires=">=3.4",
- add classifier Environment as Web Environment


vaibhavmule:
> this is likely to break the `craft info` command too
This reminded me to delete `info` file and make changes in command

josephmancuso:
fixed a few issues

PR #764, opened 2019-06-12
add SSLRedirectMiddleware

### LINKED ISSUE #741: Create an SSL redirect middleware
**Is your feature request related to a problem? Please describe.**
Need to have a way to redirect the user from `http://www.example.com` to `https://www.example.com`

**What do we currently have to do now?** 
Currently do not have anything like this.

**Describe the solution you'd like**
The middleware should be called `SSLRedirectMiddleware` and should detect if the current route is `http://` and redirect it to `https://`

vaibhavmule:
Closes #741


masonite/middleware/SSLRedirectMiddleware.py
```diff
@@ -0,0 +1,26 @@
...
+        Arguments:
+            Request {masonite.request.Request} -- The Masonite request object
+        """
+        self.request = request
+
+    def before(self):
+        """Redirect user to secure URL."""
+        from config import middleware
+        if middleware.SSLRedirect:
+            host = request.environ['HOST']
```
vaibhavmule: @josephmancuso Can I use `APP URL` as a first preference when redirecting the user to `https`?

josephmancuso: yes

masonite/middleware/SSLRedirectMiddleware.py
```diff
@@ -0,0 +1,28 @@
...
+        Arguments:
+            Request {masonite.request.Request} -- The Masonite request object
+        """
+        self.request = request
+
+    def before(self):
+        """Redirect user to secure URL."""
+        from config import middleware
+        if middleware.SSLRedirect:
+            host = application.URL or request.environ['HOST']
```
rfschubert: On this line, `request` does not exist... maybe should change to `self.request`?

vaibhavmule: Thanks, just did

josephmancuso:
This isn't really working at all. Can't get this to work on localhost. I tried to make improvements but still couldn't get it. This is something that should be added to any project by just adding a middleware but it broke the browser for me.

Will reopen if there is something that can be fixed. Trying to clean up Masonite a bit

PR #766, opened 2019-06-12
Feature/730 Add dev command

hammacktony:
Have the ability to test, lint, and clean a repo.

References #730 and #617


masonite/commands/DevCommand.py
```diff
@@ -0,0 +1,39 @@
...
+            subprocess.call(['flake8 {location} --ignore=E501,F401,E128,E402,E731,F821,E712,W503'.format(location=location)], shell=True)
+
+        # Clean repo (Works for Windows, MacOS, and Linux)
+        if self.option('clean'):
+            subprocess.call(["python3", "-c", "import pathlib; [p.unlink() for p in pathlib.Path('.').rglob('*.py[co]')]"])
+            subprocess.call(["python3", "-c", "import pathlib; [p.rmdir() for p in pathlib.Path('.').rglob('__pycache__')]"])
+
+        # Command did nothing
+        if not self.option('test') and not self.option('lint') and not self.option('clean'):
+            print('No actions selected')
```
josephmancuso: get rid of the if statement here on the bottom, use return statements at the end of each if statement and then don't use `print`. Use `self.info` or related commands. Its prettier.

hammacktony: Actually, I think there should be the third conditional. The dev can do a combination of all three things. (i.e. `craft dev --test tests/ --clean`) I did not want them to run `craft dev`, and expect for things to happen. Early returns will hinder this command.

josephmancuso: ok fair point.

vaibhavmule:
If am not wrong, isn't it implemented with new `makefile` expect the clean part?

hammacktony:
This craft command is useful for both Windows and Linux/MacOS. `make` is not native to Windows. Also, the `find` command is not the same as Linux/MacOS.

josephmancuso:
@vaibhavmule so yes and no. The way @hammacktony has built it, it will be able to be used on all masonite apps. All new 2.2 apps will have the dev command. The make file can only be used when contributing inside the core repo

vaibhavmule:
@josephmancuso if it is for masonite apps, it makes sense but the command can be simplers, as `craft test`, `craft clean`. these are a generic command, but `craft lint`, is based on user preference about what he wants to ignore or keep, which is dynamic?

hammacktony:
For windows devs, `make` is not native and a bit of a hassle to install.

vaibhavmule:
@hammacktony  Windows has Windows Subsystem for Linux, can't I use makefile with that?

hammacktony:
Not all people use WSL. I have a coworker who doesn't. Percentage wise, more people use Python Windows machines than MacOS and Linux. I know? Weird. Many of them do not use WSL.

hammacktony:
I am going to split this up into `dev:test`, `dev:lint`, `dev:clean`. Cleaner api.

josephmancuso:
hmm i think splitting it up is a bad idea. I liek the idea of being able to do multiple at once and having separate commands defeats that goal.

hammacktony:
Fair enough. Didn't want to do an anti-pattern.

PR #801, opened 2019-06-25
added more dynamic testing

josephmancuso:


josephmancuso:
still wont really work well and its getting a bit complex

PR #837, opened 2019-07-08
Remove a duplicated routing rule

Leon0824:


josephmancuso:
awesome. Didn't notice :)

PR #863, opened 2019-08-03
add storage/uploads to .gitignore

TRSridhar:
Added the line storage/uploads to this file


vaibhavmule:
Thanks, @TRSridhar for PR, I will keep this open for @josephmancuso to look on requirement is `/storage/uploads` folder can be committed but not the files of it.

PR #867, opened 2019-08-03
Fix dangerous arguments

manishpriyap:


vaibhavmule:
@manishpriyap Thanks for the PR, just want to know how will it help?

masonite/commands/ServeCommand.py
```diff
@@ -46,7 +46,9 @@ def handle(self):
             from ._devserver import run
             run(self.option("host"), self.option("port"), application)
 
-    def _run_reloader(self, reloader, extra_files=[]):
+    def _run_reloader(self, reloader, extra_files=None):
```
vaibhavmule: just curious to know how these arguments can be dangerous?

Issue #727, opened 2019-05-27
Optimize Route Provider


hammacktony:
In response to Masonite's rankings in [https://github.com/vaibhavmule/web-frameworks](https://github.com/vaibhavmule/web-frameworks), we should aim to optimize the Route Provider.

Currently, with one run, Masonite processes a request at 40 microseconds. I believe we can beat that time. I propose using a Radix Trie structure (or any Trie structure) for http routing than a simple for-loop. 

There are two packages I have found that could be inspiration for us. One is [xrtr](https://xrtr.readthedocs.io/en/latest/index.html) (Cython) and the other is [Kua](https://kua.readthedocs.io/en/latest/api.html) (Pure-Python). I believe we can create our own package based on these than achieve faster times than the current model.

Here is a quick benchmark based on processing a single route: `xrtr: 948 ns, kua: 13.1 micro secs, masonite: 40 microseconds`

So, Masonite is 42 times slow than xrtr, and Kua is 14x slower than xrtr. 

This should be targeted for a 2.3 release.

PR #890, opened 2019-09-03
Feature/727

hammacktony:
In reference to #727, optimizing the route provider. This PR includes a tree based router instead of the formal for-loop.

Also, route compiling works now.


hammacktony:
Still have to work on `test_route_provider.py`.

PR #898, opened 2019-09-12
reflector run_middleware

erhuabushuo:


erhuabushuo:
> What does this change do exactly? Just cleans up that code?

Yes, more efficiently，easy to understand

josephmancuso:
Ok well tests looks like they pass so LGTM!

PR #905, opened 2019-09-21
updated exception_handler.py to return json response for Accept header

garyburgmann:
updated exception_handler.py to return a json response if the client makes a request with the 'Accept: application/json' header


josephmancuso:
LGTM ! 👍

PR #907, opened 2019-09-22
Adding old helper function to recover values from submitted forms

Marlysson:
@josephmancuso Seems that until now are everything fine.

Related to test to this feature. I need create the whole environment with: Set some data to request, flash them to session and after validate them with Validators and verify if the data recover by new **old** function return the correct data?


josephmancuso:
looks awesome :) made just a few changes

PR #910, opened 2019-09-28
directly returning a Paginator or LengthAwarePaginator instance

garyburgmann:
directly returning a Paginator or LengthAwarePaginator instance from a controller now creates a JSON response. The JSON response includes the collection data as well as the meta data regarding paginated collection.

### DIFF
--- a/masonite/response.py
+++ b/masonite/response.py
@@ -8,6 +8,8 @@ from masonite.view import View
 
 from orator.support.collection import Collection
 from orator import Model
+from orator.pagination.length_aware_paginator import LengthAwarePaginator
+from orator.pagination.paginator import Paginator
 from masonite.app import App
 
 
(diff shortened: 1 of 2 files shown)


garyburgmann:
Have also made a pull req. into docs, and nested the info in the basics/controllers section

josephmancuso:
This is perfect. Can you just add a simple test somewhere to assert that the returned JSON has a few paginated results? This way we can track if it breaks in the future.

garyburgmann:
Will do. Under the pump at the moment, but hopefully will find an opportunity in the next 48 hours

garyburgmann:
tests written for this now @josephmancuso

josephmancuso:
i merged this into master instead of 2.2 4b7c4c9685a6b65afc28d4ff30619ba2080640b3

PR #945, opened 2020-01-04
Feature/944

Marlysson:
Trying add some server to handle Option preflight requests.

 - [ ] Add section on documentation


Marlysson:
@josephmancuso When I ran the local server, the following message showed:

>The Masonite development server is not capable of handling OPTIONS preflight requests.
You should use a more powerful server if using the Option

But I wasn't able to find where this message was defined, can you show me?

And to test the new server's behavior... Would I create a new project from this new masonite changed structure? Because I saw that normal project's flow is running from site-package > masonite-cli ( create projects, run migrations, run server )

